PR #5, opened 2018-11-20
bpf: sync with latest bpf-next tree

yonghong-song:
Syncing is up to the commit:

sync with latest bpf-next tree.
the include/linux/filter.h is created as libbpf.c tries
to use various insn define macros.

Signed-off-by: Yonghong Song <yhs@fb.com>


yonghong-song:
@rdna Could you help test whether the change is good or not in FB environment? I tested the library build (static, shared) is fine, but not sure whether we have any malfunction or not during linking...

src/bpf.c
```diff
@@ -196,19 +206,66 @@ int bpf_load_program_xattr(const struct bpf_load_program_attr *load_attr,
 	attr.log_level = 0;
 	attr.kern_version = load_attr->kern_version;
 	attr.prog_ifindex = load_attr->prog_ifindex;
+	attr.prog_btf_fd = load_attr->prog_btf_fd;
```
rdna: This seems to break loading BPF progs.

yonghong-song:
tested and worked fine with facebook internal integration.

PR #11, opened 2019-02-15
sync with latest bpf-next

anakryiko:
Sync latest libbpf sources. Tested against pahole.


yonghong-song:
thanks!

PR #20, opened 2019-03-28
Build pkg-config file, use CPPFLAGS

bluca:
As promised on the LKML, here's the changes to build the pc file in this project too. This PR depends on a new sync from bpf-next that will bring in the libbpf.pc.template file added by commit:

https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/?id=dd399ac9e343c7573c47d6820e4a23013c54749d

And pass CPPFLAGS to the compiler.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -63,7 +63,7 @@ $(OBJDIR)/libbpf.pc:
 		< libbpf.pc.template > $@
 
 $(OBJDIR)/%.o: %.c
-	$(CC) $(ALL_CFLAGS) -c $< -o $@
+	$(CC) $(ALL_CFLAGS) $(CPPFLAGS) -c $< -o $@
 
 define do_install
 	if [ ! -d '$(DESTDIR)$2' ]; then		\
--- a/src/Makefile
+++ b/src/Makefile
@@ -31,6 +31,8 @@ HEADERS := bpf.h libbpf.h btf.h
 UAPI_HEADERS := $(addprefix $(TOPDIR)/include/uapi/linux/,bpf.h bpf_common.h \
 	btf.h)
 
+PC_FILE := libbpf.pc
+
 INSTALL = install
 
 DESTDIR ?=
@@ -46,13 +48,19 @@ LIBDIR ?= $(PREFIX)/$(LIBSUBDIR)
 INCLUDEDIR ?= $(PREFIX)/include
 UAPIDIR ?= $(PREFIX)/include
 
-all: $(LIBS)
+all: $(LIBS) $(PC_FILE)
 
 $(OBJDIR)/libbpf.a: $(OBJS)
 	$(AR) rcs $@ $^
 
 $(OBJDIR)/libbpf.so: $(OBJS)
 	$(CC) -shared $(LDFLAGS) $^ -o $@
+
+$(OBJDIR)/libbpf.pc:
+	sed -e "s|@PREFIX@|$(PREFIX)|" \
+		-e "s|@LIBDIR@|$(LIBDIR)|" \
+		-e "s|@VERSION@|$(LIBBPF_VERSION)|" \
+		< libbpf.pc.template > $@
 
 $(OBJDIR)/%.o: %.c
 	$(CC) $(ALL_CFLAGS) -c $< -o $@
@@ -64,7 +72,7 @@ define do_install
 	$(INSTALL) $1 $(if $3,-m $3,) '$(DESTDIR)$2'
 endef
 
-install: all install_headers
+install: all install_headers install_pkgconfig
 	$(call do_install,$(LIBS),$(LIBDIR))
 
 install_headers:
@@ -75,5 +83,8 @@ install_headers:
 install_uapi_headers:
 	$(call do_install,$(UAPI_HEADERS),$(UAPIDIR)/linux,644)
 
+install_pkgconfig: $(PC_FILE)
+	$(call do_install,$(PC_FILE),$(LIBDIR)/pkgconfig,644)
+
 clean:
-	rm -f *.o *.a *.so
+	rm -f *.o *.a *.so *.pc


bluca:
Note that there's no version defined in this repository for now, so that field will be empty.

src/Makefile
```diff
@@ -16,6 +16,8 @@ endif
 
 CFLAGS ?= -g -O2 -Werror -Wall
 ALL_CFLAGS += $(CFLAGS)
+ALL_CFLAGS += $(shell command -v pkg-config > /dev/null 2>&1 && pkg-config --cflags libelf)
+LDFLAGS += $(shell command -v pkg-config > /dev/null 2>&1 && pkg-config --libs libelf || echo "-lelf")
```
rdna: Please don't modify `ALL_CFLAGS` and `LDFLAGS` in this way. It should contain only flags that are absolutely necessary for everyone.
All use case specific flags should be customized via specifying `CFLAGS` or `LDFLAGS` by build system.

One of the pain points of e.g. bringing libbpf from kernel-tree to Facebook infra was the one that kernel's libbpf was adding tons of flags that are irrelevant or even breaking our build system.

That's why this Makefile was written in a way that it provides by default only absolutely necessary minimum of flags and all others is build-system's business.

bluca: I understand your concern, but as far as I can see libelf is not optional, the library is using symbols from it without ifdefs or anything. Am I missing something?

rdna: * pkg-config is optional;
* there are at least two ways to satisfy elf dependency: `libelf` and `elfutils` and I know at least one build system that uses `elfutils` for `libbpf`

bluca: Usage of pkg-config here is optional as well - as you can see, if it's not present it will gracefully fall back to the hard-coded -lelf flag.

Distros use elfutils as far as I can see, which ships the pkg-config file libelf.pc. By libelf, which project do you refer to? If it ships a libelf.so, this will work as well as it's the same flag.

rdna: It is _not_ optional and this is the problem.

Presence of `pkg-config` and `libelf` on a system doesn't mean that build system wants to use them. Build system may have its own repo with its own paths, library versions, patches to e.g. `elfutils` sources. But with this change one would required to make sure that either `pkg-config` or `libelf` are not present in building environment. It can lead to non-deterministic build results.

By `libelf` I refer to this ancient stuff: https://directory.fsf.org/wiki/Libelf.

anakryiko: @rdna, do you have specific suggestions how this Makefile could look like to support building both for pkg-config and w/o it?

@bluca, I don't know much about pkg-config, but would it be acceptable to have either a separate target or extra make flag for pkg-config, which would trigger adding extra CFLAGS/LDFLAGS only if pkg-config is desired?

bluca: That old library ships the exact same libelf.pc and libelf.so that elfutils does, so this change is compatible with both. Given they have the same names, only one can be installed in the canonical system path, as they conflict, so there's no non-determinism.

Most importantly, pkg-config is the normal and most common way of handling such build environment changes - one defines a preferred PKG_CONFIG_PATH environment variable that drives pkg-config toward one installation or the other, deterministically and most importantly without any special sauce specific to this or that project. It makes things like cross compilation saner.

Eg:

```
$ cd /tmp/libelf-0.8.13
$ ./configure --prefix=/tmp/elf; make; make install
loading cache ./config.cache
<...>
$ cd ~/git/libbpf/src
$ PKG_CONFIG_PATH=/tmp/elf/lib/pkgconfig make BUILD_SHARED=1
ccache gcc -I. -I../include -I../include/uapi -fPIC -fvisibility=hidden -g -O2 -Werror -Wall -I/tmp/elf/include/libelf -I/tmp/elf/include  -c bpf.c -o bpf.o
ccache gcc -I. -I../include -I../include/uapi -fPIC -fvisibility=hidden -g -O2 -Werror -Wall -I/tmp/elf/include/libelf -I/tmp/elf/include  -c btf.c -o btf.o
ccache gcc -I. -I../include -I../include/uapi -fPIC -fvisibility=hidden -g -O2 -Werror -Wall -I/tmp/elf/include/libelf -I/tmp/elf/include  -c libbpf.c -o libbpf.o
ccache gcc -I. -I../include -I../include/uapi -fPIC -fvisibility=hidden -g -O2 -Werror -Wall -I/tmp/elf/include/libelf -I/tmp/elf/include  -c libbpf_errno.c -o libbpf_errno.o
ccache gcc -I. -I../include -I../include/uapi -fPIC -fvisibility=hidden -g -O2 -Werror -Wall -I/tmp/elf/include/libelf -I/tmp/elf/include  -c netlink.c -o netlink.o
ccache gcc -I. -I../include -I../include/uapi -fPIC -fvisibility=hidden -g -O2 -Werror -Wall -I/tmp/elf/include/libelf -I/tmp/elf/include  -c nlattr.c -o nlattr.o
ccache gcc -I. -I../include -I../include/uapi -fPIC -fvisibility=hidden -g -O2 -Werror -Wall -I/tmp/elf/include/libelf -I/tmp/elf/include  -c str_error.c -o str_error.o
ccache gcc -I. -I../include -I../include/uapi -fPIC -fvisibility=hidden -g -O2 -Werror -Wall -I/tmp/elf/include/libelf -I/tmp/elf/include  -c libbpf_probes.c -o libbpf_probes.o
ccache gcc -I. -I../include -I../include/uapi -fPIC -fvisibility=hidden -g -O2 -Werror -Wall -I/tmp/elf/include/libelf -I/tmp/elf/include  -c bpf_prog_linfo.c -o bpf_prog_linfo.o
ccache gcc -I. -I../include -I../include/uapi -fPIC -fvisibility=hidden -g -O2 -Werror -Wall -I/tmp/elf/include/libelf -I/tmp/elf/include  -c xsk.c -o xsk.o
ar rcs libbpf.a bpf.o btf.o libbpf.o libbpf_errno.o netlink.o nlattr.o str_error.o libbpf_probes.o bpf_prog_linfo.o xsk.o
ccache gcc -shared -L/tmp/elf/lib -lelf bpf.o btf.o libbpf.o libbpf_errno.o netlink.o nlattr.o str_error.o libbpf_probes.o bpf_prog_linfo.o xsk.o -o libbpf.so
sed -e "s|@PREFIX@|/usr|" \
	-e "s|@LIBDIR@|/usr/lib64|" \
	-e "s|@VERSION@||" \
	< libbpf.pc.template > libbpf.pc
$ ldd libbpf.so 
	linux-vdso.so.1 (0x00007fff1958e000)
	libelf.so.0 => not found
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa6ca2d7000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fa6ca4d9000)
$ LD_LIBRARY_PATH=/tmp/elf/lib/ ldd libbpf.so 
	linux-vdso.so.1 (0x00007ffcf57d8000)
	libelf.so.0 => /tmp/elf/lib/libelf.so.0 (0x00007fc7a5497000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc7a52af000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc7a54c8000)
```

bluca: > @bluca, I don't know much about pkg-config, but would it be acceptable to have either a separate target or extra make flag for pkg-config, which would trigger adding extra CFLAGS/LDFLAGS only if pkg-config is desired?

If an option is desired, IMHO it should be the opposite - do the expected, default, standard thing by default, and then add options for strange corner cases.

rdna: > @rdna, do you have specific suggestions how this Makefile could look like to support building both for pkg-config and w/o it?

@anakryiko I made specific suggestions in my first comment -- just pass `CFLAGS` from build system. No changes needed at all to achieve what @bluca wants to achieve.

bluca: @anakryiko @rdna since there's no point in keeping the other changes back and this one is independent, I've dropped it from this PR. Will reopen it later in a different PR by itself.

Issue #28, opened 2019-04-06
bpf_create_map seems to be failing under UBSan


evverx:
It was originally discovered by the systemd CI in systemd/systemd#12151:
```sh
../src/libbpf/bpf_next/src/libbpf/src/bpf.c:92:2: runtime error: null pointer passed as argument 2, which is declared to never be null
    #0 0x558b1dea3ecc in bpf_create_map_xattr ../src/libbpf/bpf_next/src/libbpf/src/bpf.c:92
    #1 0x558b1e2021db in bpf_create_map ../src/libbpf/bpf_next/src/libbpf/src/bpf.c:135
    #2 0x558b1e1ba512 in bpf_firewall_supported ../src/core/bpf-firewall.c:685
    #3 0x558b1df96a74 in cg_bpf_mask_supported ../src/core/cgroup.c:2531
    #4 0x558b1df9a04c in manager_setup_cgroup ../src/core/cgroup.c:2685
    #5 0x558b1deb044d in manager_new ../src/core/manager.c:835
    #6 0x558b1dea48a0 in main ../src/test/test-sched-prio.c:31
    #7 0x7f99c325d09a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2409a)
    #8 0x558b1dea4089 in _start (/build/build/test-sched-prio+0x496089)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/libbpf/bpf_next/src/libbpf/src/bpf.c:92:2 in
-------
```

yonghong-song:
This needs to be fixed in kernel. Could you submit a patch for this? The fix could be
```
diff --git a/src/bpf.c b/src/bpf.c
index 9cd0155..eab2fdc 100644
--- a/src/bpf.c
+++ b/src/bpf.c
@@ -79,7 +79,6 @@ static inline int sys_bpf_prog_load(union bpf_attr *attr, unsigned int size)
 
 int bpf_create_map_xattr(const struct bpf_create_map_attr *create_attr)
 {
-       __u32 name_len = create_attr->name ? strlen(create_attr->name) : 0;
        union bpf_attr attr;
 
        memset(&attr, '\0', sizeof(attr));
@@ -89,8 +88,10 @@ int bpf_create_map_xattr(const struct bpf_create_map_attr *create_attr)
        attr.value_size = create_attr->value_size;
        attr.max_entries = create_attr->max_entries;
        attr.map_flags = create_attr->map_flags;
[...]

Issue #27, opened 2019-04-06
clang seems to be failing to compile libbpf


evverx:
It was originally discovered by the systemd CI in https://github.com/systemd/systemd/pull/12151:
```sh
FAILED: src/libbpf/9e6d599@@bpf@sta/bpf_next_src_libbpf_src_libbpf.c.o
clang -Isrc/libbpf/9e6d599@@bpf@sta -Isrc/libbpf -I../src/libbpf -I../src/libbpf/bpf_next/src/libbpf/include -I../src/libbpf/bpf_next/src/libbpf/include/uapi -Xclang -fcolor-diagnostics -fsanitize=address,undefined -fno-omit-frame-pointer -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Werror -std=gnu99 -g -Wextra -Werror=undef -Wmissing-include-dirs -Wold-style-definition -Wpointer-arith -Winit-self -Wfloat-equal -Werror=missing-prototypes -Werror=implicit-function-declaration -Werror=missing-declarations -Werror=return-type -Werror=incompatible-pointer-types -Werror=format=2 -Wstrict-prototypes -Wredundant-decls -Wmissing-noreturn -Wshadow -Wendif-labels -Wstrict-aliasing=2 -Wwrite-strings -Werror=overflow -Werror=shift-count-overflow -Wdate-time -Wnested-externs -Wno-unused-parameter -Wno-missing-field-initializers -Wno-unused-result '-Wno-error=#warnings' -Wno-string-plus-int -Wno-error=nonnull -ffast-math -fno-common -fdiagnostics-show-option -fno-strict-aliasing -fvisibility=hidden -fstack-protector -fstack-protector-strong --param=ssp-buffer-size=4 -Wno-typedef-redefinition -Wno-gnu-variable-sized-type-not-at-end -Werror=shadow -include config.h -fPIC -g -O2 -Werror -Wall -Wno-pointer-arith -Wno-sign-compare  -MD -MQ 'src/libbpf/9e6d599@@bpf@sta/bpf_next_src_libbpf_src_libbpf.c.o' -MF 'src/libbpf/9e6d599@@bpf@sta/bpf_next_src_libbpf_src_libbpf.c.o.d' -o 'src/libbpf/9e6d599@@bpf@sta/bpf_next_src_libbpf_src_libbpf.c.o' -c ../src/libbpf/bpf_next/src/libbpf/src/libbpf.c
[...]

PR #36, opened 2019-04-17
Introduce Travis CI tests

wat-ze-hex:
In order to `libbpf` to be used in systemd some testing is required, see related
discussions in https://github.com/systemd/systemd/pull/12151 and
https://github.com/libbpf/libbpf/pull/29
Having CI will help track issues like https://github.com/libbpf/libbpf/issues/27 and https://github.com/libbpf/libbpf/issues/28
Tests introduced here mirror the tests of systemd:
For Debian: use `docker` virtualization, build with gcc, gcc + asan, clang, clang + asan.
Fror Ubuntu Xenial: build with gcc.

Differences:
For `libbpf` only `libelf` and its dependencies are installed.
Instead of Meson build system `make` is used, so `make` remains the preferred
method of building and `meson.build` doesn't get rooted in `libbpf`.

`travis_wait.bash` is kept as a workaround for
https://github.com/travis-ci/travis-ci/issues/9979

An example of testing UI: https://travis-ci.org/wat-ze-hex/libbpf

### DIFF
--- a/travis-ci/managers/debian.sh
+++ b/travis-ci/managers/debian.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+PHASES=(${@:-SETUP RUN RUN_ASAN CLEANUP})
+DEBIAN_RELEASE="${DEBIAN_RELEASE:-testing}"
+CONT_NAME="${CONT_NAME:-debian-$DEBIAN_RELEASE-$RANDOM}"
+DOCKER_EXEC="${DOCKER_EXEC:-docker exec -it $CONT_NAME}"
+DOCKER_RUN="${DOCKER_RUN:-docker run}"
+REPO_ROOT="${REPO_ROOT:-$PWD}"
+ADDITIONAL_DEPS=(clang)
+
+function info() {
+    echo -e "\033[33;1m$1\033[0m"
+}
+
+set -e
+
+source "$(dirname $0)/travis_wait.bash"
+
+for phase in "${PHASES[@]}"; do
+    case $phase in
+        SETUP)
+            info "Setup phase"
+            info "Using Debian $DEBIAN_RELEASE"
+	    docker pull debian:$DEBIAN_RELEASE
+            info "Starting container $CONT_NAME"
+            $DOCKER_RUN -v $REPO_ROOT:/build:rw \
+                        -w /build --privileged=true --name $CONT_NAME \
+			-dit --net=host debian:$DEBIAN_RELEASE /bin/bash
+            $DOCKER_EXEC bash -c "echo deb-src http://deb.debian.org/debian $DEBIAN_RELEASE main >>/etc/apt/sources.list"
+            $DOCKER_EXEC apt-get -y update
+            $DOCKER_EXEC apt-get -y build-dep libelf-dev
+            $DOCKER_EXEC apt-get -y install libelf-dev
+            $DOCKER_EXEC apt-get -y install "${ADDITIONAL_DEPS[@]}"
+            ;;
+        RUN|RUN_CLANG)
+            if [[ "$phase" = "RUN_CLANG" ]]; then
+                ENV_VARS="-e CC=clang -e CXX=clang++"
+            fi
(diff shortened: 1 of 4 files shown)


yonghong-song:
Tested. It works great. https://travis-ci.com/yonghong-song/libbpf/builds/108755639
Thanks!

travis-ci/managers/debian.sh
```diff
@@ -0,0 +1,60 @@
...
+            docker exec $ENV_VARS -it $CONT_NAME mkdir build
+            $DOCKER_EXEC make CFLAGS='-Werror' -C ./src -B OBJDIR=../build
+            $DOCKER_EXEC rm -rf build
+            ;;
+        RUN_ASAN|RUN_CLANG_ASAN)
+            if [[ "$phase" = "RUN_CLANG_ASAN" ]]; then
+                ENV_VARS="-e CC=clang -e CXX=clang++"
+            fi
+            docker exec $ENV_VARS -it $CONT_NAME mkdir build
+            $DOCKER_EXEC make CFLAGS='-Werror -Db_sanitize=address,undefined' -C ./src -B OBJDIR=../build
```
evverx: @wat-ze-hex thank you for bringing Travis CI to the libbpf repository! 

`-Db_sanitize` is [a `meson`-ism](https://mesonbuild.com/howtox.html#use-address-sanitizer). To build `libbpf` with `ASan+UBSan` using `make` it seems that something like `-fsanitize=address,undefined` should be used instead.

It also looks like that CFLAGS passed via the command line overrides whatever is set in `Makefile` (including `-Wall`), which I think makes `-Werror` a little bit less useful.

`ENV_VARS` don't seem to affect `make` in the sense that `make` runs `gcc` even when `phase` is `RUN_CLANG*`. It would probably be better to use something like `docker exec $ENV_VARS make CFLAGS='...' -C ./src -B OBJDIR=../build`.

And, lastly, to judge from the logs `make` complained that "make: pkg-config: Command not found" but I'm not sure how critical this is.

yonghong-song: @wat-ze-hex could you have a followup on this?

rdna: @evverx great catch!

`Makefile` from libbpf works with env vars just fine, e.g.
```
14:43:46 1 rdna@:~/.../git/libbpf$>make -C src/ -n | head -n 3
make: Entering directory `/data/users/rdna/git/libbpf/src'
cc -I. -I../include -I../include/uapi -DCOMPAT_NEED_REALLOCARRAY -fPIC -fvisibility=hidden -g -O2 -Werror -Wall    -c bpf.c -o bpf.o
cc -I. -I../include -I../include/uapi -DCOMPAT_NEED_REALLOCARRAY -fPIC -fvisibility=hidden -g -O2 -Werror -Wall    -c btf.c -o btf.o
14:43:48 0 rdna@:~/.../git/libbpf$>env CC=~/bin/clang make -C src/ -n | head -n 3
make: Entering directory `/data/users/rdna/git/libbpf/src'
/home/rdna/bin/clang -I. -I../include -I../include/uapi -DCOMPAT_NEED_REALLOCARRAY -fPIC -fvisibility=hidden -g -O2 -Werror -Wall    -c bpf.c -o bpf.o
/home/rdna/bin/clang -I. -I../include -I../include/uapi -DCOMPAT_NEED_REALLOCARRAY -fPIC -fvisibility=hidden -g -O2 -Werror -Wall    -c btf.c -o btf.o
```
The problem here is that `$ENV_VARS` is passed  to `mkdir` on line 47, but it should be passed to actual `make` on line 48 instead. That should be fixed indeed.

As for `CFLAGS` it was made this way intentionally so that user can fully override flags that are not required for the build. See a82a66eda3b1965d786960b61c1899011a15368d for details. So yeah, `-Werror` and other necessary flags should be passed explicitly.

wat-ze-hex: @evverx

Thanks, created a follow-up PR!

PR #40, opened 2019-04-19
Travis CI follow-ups

wat-ze-hex:
Follow up of PR https://github.com/libbpf/libbpf/pull/36
Fix bugs pointed by at https://github.com/libbpf/libbpf/pull/36#discussion_r276522335
Show Travis build status at README webpage.

Tested by Travis working on forked `libbpf`: https://travis-ci.org/wat-ze-hex/libbpf

### DIFF
--- a/travis-ci/managers/debian.sh
+++ b/travis-ci/managers/debian.sh
@@ -3,13 +3,18 @@
 PHASES=(${@:-SETUP RUN RUN_ASAN CLEANUP})
 DEBIAN_RELEASE="${DEBIAN_RELEASE:-testing}"
 CONT_NAME="${CONT_NAME:-debian-$DEBIAN_RELEASE-$RANDOM}"
-DOCKER_EXEC="${DOCKER_EXEC:-docker exec -it $CONT_NAME}"
+ENV_VARS="${ENV_VARS:-}"
 DOCKER_RUN="${DOCKER_RUN:-docker run}"
 REPO_ROOT="${REPO_ROOT:-$PWD}"
-ADDITIONAL_DEPS=(clang)
+ADDITIONAL_DEPS=(clang pkg-config)
+CFLAGS="-g -O2 -Werror -Wall"
 
 function info() {
     echo -e "\033[33;1m$1\033[0m"
+}
+
+function docker_exec() {
+    docker exec $ENV_VARS -it $CONT_NAME "$@"
 }
 
 set -e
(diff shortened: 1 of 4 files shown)


travis-ci/managers/debian.sh
```diff
@@ -36,16 +37,17 @@ for phase in "${PHASES[@]}"; do
             if [[ "$phase" = "RUN_CLANG" ]]; then
                 ENV_VARS="-e CC=clang -e CXX=clang++"
             fi
-            docker exec $ENV_VARS -it $CONT_NAME mkdir build
-            $DOCKER_EXEC make CFLAGS='-Werror' -C ./src -B OBJDIR=../build
+            $DOCKER_EXEC mkdir build
+            docker exec $ENV_VARS -it $CONT_NAME make CFLAGS="${CFLAGS}" -C ./src -B OBJDIR=../build
```
rdna: Nit: any reason not to use `$DOCKER_EXEC $ENV_VARS` on line 41?

travis-ci/managers/debian.sh
```diff
@@ -36,16 +37,17 @@ for phase in "${PHASES[@]}"; do
...
             ;;
         RUN_ASAN|RUN_CLANG_ASAN)
             if [[ "$phase" = "RUN_CLANG_ASAN" ]]; then
                 ENV_VARS="-e CC=clang -e CXX=clang++"
             fi
-            docker exec $ENV_VARS -it $CONT_NAME mkdir build
-            $DOCKER_EXEC make CFLAGS='-Werror -Db_sanitize=address,undefined' -C ./src -B OBJDIR=../build
+            CFLAGS="${CFLAGS} -fsanitize=address,undefined"
+            $DOCKER_EXEC mkdir build
+            docker exec $ENV_VARS -it $CONT_NAME make CFLAGS="${CFLAGS}" -C ./src -B OBJDIR=../build
```
rdna: ditto

travis-ci/managers/debian.sh
wat-ze-hex: Need to slightly modify `DOCKER_EXEC` var for that:
`
DOCKER_EXEC="${DOCKER_EXEC:-docker exec -it $CONT_NAME}"
`
vs.
`docker exec $ENV_VARS -it $CONT_NAME`

`$DOCKER_EXEC $ENV_VARS` concatenation would have another meaning.

wat-ze-hex:
@rdna 

> Also do you know how to identify compiler version travis uses for its builds? 

Well, by adding e.g. `exec $CC -v` to the scripts :D I'll do that.

> It can be useful in case it catches an error that has to be reproduced later (e.g. like it was with gcc 8 recently). It's not a blocking thing though and we can follow-up on it some time later.

Looks like a use case for another test. We should be able to fetch gcc 8 with its dependencies and use by setting `CC` env var.

travis-ci/managers/debian.sh
rdna: That's what I was trying to figure out :) OK, if order of arguments matters here, then it should be fine to go as is. Thanks!

wat-ze-hex:
I'll add gcc 8 test as the 3rd commit to this PR. Don't merge it yet.

evverx:
@wat-ze-hex thank you!

> Well, by adding e.g. exec $CC -v to the scripts and looking into logs :D I'll do that.

It seems that you're going to add another command anyway and given that `meson` shows which compiler it has found, I thought maybe it would be better to also build `libbpf` with `meson` using something like `sh -c 'meson ./build && ninja -C ./build`.

Speaking of `meson`, I registered `libbpf` on Coverity Scan and it turns out that it's somehow more compatible with Coverity than `make`. The 5 eliminated "defects" at https://scan.coverity.com/projects/libbpf were gone when I used `meson` to build `libbpf` with `cov-build`.

By the way, @wat-ze-hex @rdna would it be OK if I sent invitations (so that you can view the defects and possibly integrate TravisCI with Coverity) to the email addresses you use in the commit messages?

> We should be able to fetch gcc 8 with its dependencies and use it by setting CC env var.

In general I think it's a good idea to try to compile `libbpf` with as many compilers as possible but in my experience it's kind of hard to maintain and usually not worth it. I'd just use the compiler shipped on Debian for the time being.

rdna:
@evverx yeah, I had same thought on using `meson`. Another thing we can do is to compare `meson`'s and regular `make`'s results e.g. like those `diff`-s in https://github.com/libbpf/libbpf/pull/29#discussion_r274134966 -- this way it's easy to make sure there are no discrepancies between two ways to build.

 > would it be OK if I sent invitations (so that you can view the defects and possibly integrate TravisCI with Coverity) to the email addresses you use in the commit messages?

Sure, sounds great!

> In general I think it's a good idea to try to compile libbpf with as many compilers as possible but in my experience it's kind of hard to maintain and usually not worth it. I'd just use the compiler shipped on Debian for the time being.

If this is fine from systemd integration PoV then I'm fine with this as well :)

evverx:
@rdna I noticed you took a look at the Coverity report and even marked one defect as a false positive, which is great and I'll get back to the false positive later :-)

@wat-ze-hex @rdna I just wanted to say that I invited you as onwers/maintainers, which means that you can do anything there including removing me from there if you'll find my lurking there annoying :-)

Regarding [the false positive](https://scan3.coverity.com/reports.htm#v37540/p11903/fileInstanceId=25574504&defectInstanceId=7657707&mergedDefectId=203822&eventId=7657707-8), I think it's true that in most cases the file descriptor passed to `dup3` is closed but judging by https://github.com/torvalds/linux/blob/38e7571c07be01f9f19b355a9306a4e3d5cb0f5b/fs/file.c#L909-L934, there are places where something can go wrong before O_CLOEXEC is taken into account so just to be safe I'd assume that `new_fd` is closed only when `dup3` doesn't return -1.

evverx:
Apparently I was thinking about something else because `O_CLOEXEC` has nothing to do with what I was trying to say :-) My point was that it seems to be possible for `new_fd` to be open after a call to `dup3` if it hasn't been successful.

evverx:
I missed the part where `new_fd` is closed if the return value of dup3 is negative. Sorry for the noice. I'll show myself out :-)

travis-ci/managers/debian.sh
wat-ze-hex: Resolved by adding `docker_exec` function: https://github.com/libbpf/libbpf/pull/40/commits/0ba9ce75796df3fe70771d0d25af7b1162adef55#diff-6e5ee78613021bcff97487ad55a3748eR17

wat-ze-hex: Also gcc-8 tests are added for Debian: https://travis-ci.org/libbpf/libbpf/builds/526211312

travis-ci/managers/debian.sh
```diff
@@ -3,15 +3,20 @@
...
-ADDITIONAL_DEPS=(clang)
+ADDITIONAL_DEPS=(clang pkg-config)
+CFLAGS="-g -O2 -Werror -Wall"
 
 function info() {
     echo -e "\033[33;1m$1\033[0m"
 }
 
+
+function docker_exec() {
```
rdna: Do you know why `DOCKER_EXEC` was set via variable that could be overridden via env var? (`${DOCKER_EXEC:-...}`). Is it required by Travis CI to be able to override it or not really?

wat-ze-hex: @rdna `DOCKER_EXEC` is initially set in the upper level config: https://github.com/libbpf/libbpf/blob/master/.travis.yml#L74
From what I see it's not required, looks like it can also be removed from `.travis.yml`.

evverx: Just a bit of context. Historically those scripts were over-engineered. Although they were simplified a bit (by removing unnecessary steps like creating images and sending them to Docker Hub) they are still kind of opaque and I don't always understand what they are trying to do and why as in https://github.com/systemd/systemd/pull/12402#issuecomment-486765499. In my opinion most of those environment variables (including DOCKER_EXEC) can be safely removed without losing anything (and probably making the scripts easier to understand and run manually).

wat-ze-hex: `DOCKER_EXEC` is removed.

PR #41, opened 2019-05-08
Support less common environments like Alpine (musl)

acj:
This fixes the build in environments such as Alpine Linux (musl).

/cc @yonghong-song per discussion in https://github.com/iovisor/bcc/issues/2338

----

For reference, this was the build error:

```
# make
[...]
cc -I. -I../include -I../include/uapi -DCOMPAT_NEED_REALLOCARRAY -g -O2 -Werror -Wall -c libbpf.c -o libbpf.o
In file included from /usr/include/linux/byteorder/little_endian.h:13,
                 from /usr/include/asm/byteorder.h:5,
                 from /usr/include/linux/perf_event.h:20,
                 from libbpf.c:32:
/usr/include/linux/swab.h:161:8: error: unknown type name '__always_inline'
 static __always_inline __u16 __swab16p(const __u16 *p)
        ^~~~~~~~~~~~~~~
/usr/include/linux/swab.h:161:30: error: expected '=', ',', ';', 'asm' or '__attribute__' before '__swab16p'
 static __always_inline __u16 __swab16p(const __u16 *p)
                              ^~~~~~~~~
/usr/include/linux/swab.h:174:8: error: unknown type name '__always_inline'
 static __always_inline __u32 __swab32p(const __u32 *p)
        ^~~~~~~~~~~~~~~
/usr/include/linux/swab.h:174:30: error: expected '=', ',', ';', 'asm' or '__attribute__' before '__swab32p'
 static __always_inline __u32 __swab32p(const __u32 *p)
                              ^~~~~~~~~
/usr/include/linux/swab.h:187:8: error: unknown type name '__always_inline'
 static __always_inline __u64 __swab64p(const __u64 *p)
        ^~~~~~~~~~~~~~~
/usr/include/linux/swab.h:187:30: error: expected '=', ',', ';', 'asm' or '__attribute__' before '__swab64p'
 static __always_inline __u64 __swab64p(const __u64 *p)
                              ^~~~~~~~~
/usr/include/linux/swab.h:242:23: error: expected ';' before 'void'
 static __always_inline void __swab32s(__u32 *p)
                       ^~~~~
                       ;
/usr/include/linux/swab.h:255:23: error: expected ';' before 'void'
 static __always_inline void __swab64s(__u64 *p)
                       ^~~~~
                       ;
```

### DIFF
--- a/include/linux/types.h
+++ b/include/linux/types.h
@@ -7,6 +7,8 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#include <linux/stddef.h>
+
 #include <asm/types.h>
 #include <asm/posix_types.h>
 


yonghong-song:
This needs to submit to linux netdev mailing list. Could you send there? Could you also include info why other arch and/or libc do not have issues?

acj:
> Could you also include info why other arch and/or libc do not have issues?

From what I can gather, the root cause is that swab.h was changed [here](https://git.edevau.net/Ede_Vau/Linux_kernel_source_tree/commit/bc27fb68aaad) to reference `__always_inline`, but it doesn't include linux/stddef.h where `__always_inline` is defined. Building with glibc seems to avoid this issue because `__always_inline` is [conditionally redefined in sys/cdefs.h](https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/sys/cdefs.h;h=b1695376dc07872ccbb8d7dfec2e5c2975187f62;hb=HEAD#l314) which is included deep in the stdlib.h chain in my Ubuntu 18.04 builds. But that header is internal to glibc, and Alpine's musl has no equivalent workaround.

There was a [kernel patch](https://lkml.org/lkml/2018/1/3/149) submitted for the swab.h issue last year, but it didn't generate any discussion at the time and hasn't landed.

> This needs to submit to linux netdev mailing list. Could you send there?

I can, but I wonder if it would be better to maintain this as a patch for less common environments until the swab.h issue is sorted out. What do you think?

yonghong-song:
Yes, the kernel patch seems just sitting there and nobody picks it up. Probably because most people uses glibc and won't see the issue.

I guess since your issue has been resolved. We should be fine. We can have this patch sitting here as a record until swab.h issue is resolved and we do not need to do anything.

anakryiko:
@acj, is this still an issue for you or can we close this?

acj:
It's still an issue, but I'm okay with closing this PR for now.

vcunat:
Perhaps this can be fixed at some point?

Starting with Linux kernel contribution seems quite a high overhead if it's about such a trivial one-liner, so I'd hope that someone from bpf-related kernel contributors could do that.

anakryiko:
I can send a patch but I have no way to test and verify.

thefallentree:
I can add a CI to test build under alpine

anakryiko:
So actually this is Github-only change (include/linux/* files are re-implemented for Github repo), so we can land it as is.

anakryiko:
> Perhaps this can be fixed at some point?
> 
> Starting with Linux kernel contribution seems quite a high overhead if it's about such a trivial one-liner, so I'd hope that someone from bpf-related kernel contributors could do that.

I agree about the high overhead of submitting patches to libbpf, but that's unfortunate reality that libbpf's authoritative sources live in kernel repo. There isn't much realistic way around this at this point.

tamird:
I believe the intent is that libbpf is self contained - it seems that this patch broke that, as `linux/stddef.h` isn't in this repo. What would be required to add it, or otherwise fix this breakage (perhaps by making the include conditional)?

If this repo has tests, should they be using clang's `-nostdlibinc` to check against regressions of this kind?

tamird:
Actually, is this patch still needed? It looks like this PR was opened in 2019, and then merged in 2022. In the intervening period, 4c55ba2b196e1504868a253c11b7b7529b0918af pulled all the necessary definitions into libbpf. Perhaps this can be reverted without regression now?

PR #58, opened 2019-07-08
Libbpf sync 2019-07-08T18:11:08.594Z

anakryiko:
Latest and greatest stuff from libbpf. Tested as usual: make + internal FB tests.

- various bug/warning fixes
- tracing attach APIs 
- perf buffer API
- sockopt hooks support
- updated BTF-defined maps support


anakryiko:
@yonghong-song, there is still compiler warning-turned-error for GCC8 with similar strncpy issue I've fixed once (I have no idea how I missed another one, I distinctively remember searching for strncpy in xsk.c). I think we should still sync, there is a lot of useful stuff in libbpf sync, and non-GCC8 build are all passing, so we should be able to notice other issues still. And then I'll fix remaining strncpy issue in bpf tree. What do you think?

yonghong-song:
Okay, should be okay.

PR #62, opened 2019-08-01
libbpf: fix extraversion in Makefile

julia-hex:
The current LIBBPF_VERSION (0.0.3) doesn't match with ABI version LIBBPF_0.0.4.
Fix the EXTRAVERSION portion.


anakryiko:
Thanks!

PR #70, opened 2019-08-22
filter.h: fix BPF_LD_MAP_VALUE definition

petarpenkov:
The current definition calls BPF_LD_IMM64_RAW_FULL with
BPF_PSEUDO_MAP_FD but the original patch[0] invokes it with
BPF_PSEUDO_MAP_VALUE.

[0] https://patchwork.ozlabs.org/patch/1082785/


yonghong-song:
Thanks for the fix!

PR #72, opened 2019-09-25
Fix shared library build with some compilers

WOnder93:
See commit log of second commit for more details. The first commit introduces a check for this bug into Travis CI. You can verify it fails on Debian Testing and Xenial and passes after this commit. Note that there is a pre-existing build failure with ASan+UBSan due to static checker warnings.


travis-ci/managers/xenial.sh
```diff
@@ -14,4 +14,6 @@ CFLAGS="-g -O2 -Werror -Wall -fsanitize=address,undefined"
 mkdir build
 cc --version
 make CFLAGS="${CFLAGS}" -C ./src -B OBJDIR=../build
+ldd build/libbpf.so
+ldd build/libbpf.so | grep -q libelf
```
anakryiko: can you please also add helpful error message here, similar to debian.sh above?

PR #74, opened 2019-09-26
Libbpf sync 2019-09-26T20:03:17.185Z

anakryiko:
Latest fixes from bpf tree.


src/xsk.h
```diff
@@ -168,6 +168,21 @@ static inline void *xsk_umem__get_data(void *umem_area, __u64 addr)
 	return &((char *)umem_area)[addr];
 }
 
+static inline __u64 xsk_umem__extract_addr(__u64 addr)
+{
+	return addr & XSK_UNALIGNED_BUF_ADDR_MASK;
```
orivej: Since this is a public header (installed with libbpf) and `include/uapi/linux/if_xdp.h` is private (not installed with libbpd), applications (such as dpdk) can not compile with libbpf >= 0.5 and linux-headers < 5.4, e.g dpdk:
```
In file included from ../drivers/net/af_xdp/rte_eth_af_xdp.c:19:
/nix/store/b7zd8fhanxlbnp6kiqyasyvbx15wcwa4-libbpf-0.0.6/include/bpf/xsk.h: In function ‘xsk_ring_prod__needs_wakeup’:
/nix/store/b7zd8fhanxlbnp6kiqyasyvbx15wcwa4-libbpf-0.0.6/include/bpf/xsk.h:82:21: error: ‘XDP_RING_NEED_WAKEUP’ undeclared (first use in this function)
   82 |  return *r->flags & XDP_RING_NEED_WAKEUP;
```
Is this expected, or should `xsk.h` be compatible with older kernel headers?

anakryiko: @orivej, libbpf contains up to date copies of UAPI headers (including if_xdp.h) which are used to build libbpf itself. If possible, try to use those UAPI headers instead of potentially outdated system-wide headers. If not, then your only option would be updating linux-headers.

PR #79, opened 2019-10-01
Fix make install and exercise it as part of CI builds

anakryiko:


travis-ci/managers/debian.sh
```diff
@@ -30,55 +30,46 @@ for phase in "${PHASES[@]}"; do
...
-            docker_exec ldd build/libbpf.so
-            if ! docker_exec ldd build/libbpf.so | grep -q libelf; then
-                error "No reference to libelf.so in libbpf.so!"
-                exit 1
-            fi
-            docker_exec rm -rf build
-            ;;
-        RUN_ASAN|RUN_CLANG_ASAN|RUN_GCC8_ASAN)
-            if [[ "$phase" = "RUN_CLANG_ASAN" ]]; then
+            elif [[ "$phase" = "RUN_CLANG_ASAN" ]]; then
```
wat-ze-hex: `RUN_ASAN` branch is missed, so `-fsanitize=address,undefined` are not set: https://travis-ci.org/libbpf/libbpf/jobs/592133451

anakryiko: oh, great catch! fixing

travis-ci/managers/debian.sh
```diff
@@ -30,55 +30,41 @@ for phase in "${PHASES[@]}"; do
...
             docker_exec bash -c "echo deb-src http://deb.debian.org/debian $DEBIAN_RELEASE main >>/etc/apt/sources.list"
             docker_exec apt-get -y update
             docker_exec apt-get -y build-dep libelf-dev
             docker_exec apt-get -y install libelf-dev
             docker_exec apt-get -y install "${ADDITIONAL_DEPS[@]}"
             ;;
-        RUN|RUN_CLANG|RUN_GCC8)
-            if [[ "$phase" = "RUN_CLANG" ]]; then
+        RUN|RUN_CLANG|RUN_GCC8|RUN_ASAN|RUN_CLANG_ASAN|RUN_GCC8_ASAN)
+            if [[ "$phase" == *"CLANG"* ]]; then
```
wat-ze-hex: nit: let's use either `==` or `=` in `if` statement. Imo `=` is better since `==` is a bash-ism.

anakryiko: if there is no difference in behavior and it still works (I'll check) then yeah, I'll update before pushing.

Thanks for review!

anakryiko: well, it obvious does work because I accidentally used '=' for GCC8 case :)

PR #88, opened 2019-10-24
lgtm: fix the extraction process

mrc0mmand:
This PR fixes the LGTM analysis.

As this project uses only Makefile, without any configuration step, and due to a "non-standard" location of the source files, LGTM kept failing to find the respective Makefile and build the sources. By tricking LGTM's build system auto detection, that we use automake/configure, it correctly sets the source dir, thus the compilation, extraction & analysis steps now work in the `src/` subdirectory, as expected.

~~The `CFLAGS=-Wno-stringop-overflow` workaround should stay there only until #87 is resolved.~~

As mentioned in the comment below, I played around with the LGTM configuration in a separate repository, so you can see it in action before merging this PR:

 - LGTM results: https://lgtm.com/projects/g/mrc0mmand/libbpf-lgtm-test/
 - PR comments: https://github.com/mrc0mmand/libbpf-lgtm-test/pull/1#issuecomment-545898862

After merging this PR, simple build retry in the LGTM web UI should "unlock" the access to enable auto PR reviews and other useful settings.

PR #91, opened 2019-10-29
README: add a LGTM badge

mrc0mmand:
Follow-up to https://github.com/libbpf/libbpf/pull/88 (especially https://github.com/libbpf/libbpf/pull/88#issuecomment-547600163, cc @evverx)

The result looks like this:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,9 @@ Other header files at this repo (`include/linux/*.h`) are reduced versions of
 their counterpart files at bpf-next's `tools/include/linux/*.h` to make compilation
 successful.
 
-Build [![Build Status](https://travis-ci.org/libbpf/libbpf.svg?branch=master)](https://travis-ci.org/libbpf/libbpf)
+Build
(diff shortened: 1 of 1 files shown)


mrc0mmand:
Also, looks like the LGTM integration is working, according to the "checks" section!

anakryiko:
Grades A, B, C, etc are not that meaningful, IMO, should we try the one with a number of alerts instead?

mrc0mmand:
> Grades A, B, C, etc are not that meaningful, IMO, should we try the one with a number of alerts instead?

No problem, updated:

image

evverx:
I think grades are useful especially when there're a lot of "benign" alerts and alerts produced by custom queries that [LGTM kind of downgrades while rating projects](https://help.semmle.com/lgtm-enterprise/user/help/project-scoring-grading.html). By looking just at numbers, it's hard to tell what's going on. Then again, the grade is always shown at the top so it shouldn't be hard to find anyway.

PR #92, opened 2019-10-30
travis: automate Coverity builds

mrc0mmand:
Let's automate Coverity builds as well, to improve code quality. 

As the build frequency is rate limited, the Coverity runs are limited to the master branch only - that means someone with appropriate privileges has to setup a Travis CI cron job, which would run Travis once per day on the master branch, to get the optimal results.

~~Marking as WIP for now, as I've never used the Travis CI `coverity_scan` addon before, so I'd like to make sure it works as expected.~~ It appears that the `coverity_scan` addon is not as versatile as I'd like it to be, so let's fall back to the *manual* method I'm familiar with.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,19 @@ services:
 
 env:
     global:
-        - AUTHOR_EMAIL="$(git log -1 $TRAVIS_COMMIT --pretty=\"%aE\")"
+        - AUTHOR_EMAIL="$(git log -1 --pretty=\"%aE\")"
         - CI_MANAGERS="$TRAVIS_BUILD_DIR/travis-ci/managers"
         - REPO_ROOT="$TRAVIS_BUILD_DIR"
+
+stages:
+    # Run Coverity periodically instead of for each PR for following reasons:
+    # 1) Coverity jobs are heavily rate-limited
+    # 2) Due to security restrictions of encrypted environment variables
+    #    in Travis CI, pull requests made from forks can't access encrypted
+    #    env variables, making Coverity unusable
+    #    See: https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions
+    - name: Coverity
+      if: type = cron
 
 jobs:
     include:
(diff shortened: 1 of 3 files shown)


mrc0mmand:
I managed forgot, that Travis CI has a [security policy](https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions) for the encrypted environment variables. Although I can create them (as it relies on a *simple* public-key crypto), pull requests made **outside** of the original repo (like in this case), cannot use them. That's also the reason why Coverity scans can't be enabled for each PR, but has to be tied to either a cron job or a dedicated branch in the source repo.

I'll leave the PR in the state I think it will work in and will ask @anakryiko to setup a daily cron job in the Travis CI, Also, if you want, you can test it in a separate pull request, where the source branch must be from this repo.

cc @evverx

mrc0mmand:
> @mrc0mmand thank you!
> 
> According to https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase
> 
> > When one of the build commands returns a non-zero exit code, the Travis CI build runs the subsequent commands as well and accumulates the build result.

Good point. I'm not sure how I could forget that even though I kept staring directly at it, as it is "solved" in all previous jobs using `set -e/set +e`, *sigh*.

> 
> I think it would be better to put all the commands specified in the "script" section in a script with `set -e; set -x` and run just that script. This way Travis would bail out faster and more importantly it would be easier to figure out where exactly the script has failed.

Thanks to the following point a separate script is not needed anymore, as right now it's just a single command.

> 
> I'd also keep the coverity script in the repository because from time to time they have trouble updating their certificate, for example, so it's necessary to tweak the options passed to `curl`.

Definitely makes sense, updated!

> Plus, in its current form, the script just exits with 0 if Coverity Scan fails to reply. In my experience exiting with 1 there makes much more sense.

Looking at the current script it returns 0 only if we exceed the daily/weekly quota. Is that what you meant?

mrc0mmand:
Also, I found a bunch of issues when trying to run the Coverity scan locally, all of which should be fixed now:

 - use `/var/tmp` instead of `/tmp` as on systems, where `/tmp` is a `tmpfs` it may sometimes be too small to handle all Coverity data
 - `COVERITY_SCAN_BUILD_COMMAND="make -C src/"` will not work as expected - even though the compilation succeeds, Coverity will try to analyze everything from the repo instead of just files under `src/`, which leads to errors like `the percentage of compilation units ready for analysis is 48% which is less than the expected 85%`
    - fixed by splitting the build command into `COVERITY_SCAN_BUILD_COMMAND_PREPEND="cd src"` and `COVERITY_SCAN_BUILD_COMMAND="make"`

Also, I noticed a weird `Coverity Scan upload failed: Build successfully submitted..` error, but can't investigate that further right now, as I exceeded the hourly quota :-)

mrc0mmand:
> Also, I noticed a weird `Coverity Scan upload failed: Build successfully submitted..` error, but can't investigate that further right now, as I exceeded the hourly quota :-)

Apparently, this seems to be a known issue and is already fixed in the systemd repo, so I just borrowed the fix and after a quick test it indeed works.

https://github.com/libbpf/libbpf/blob/5a8cdf5c1f6e0292642dd8fd0996b4f53877aaa8/scripts/coverity.sh#L99-L101

anakryiko:
ok, let's try this, thanks!

anakryiko:
@mrc0mmand, I noticed LGTM is reporting "check failed"  because "No code changes detected". Is this expected? Can we avoid somehow this misleading message?
Screen Shot 2019-11-01 at 11 23 34 PM

mrc0mmand:
> @mrc0mmand, I noticed LGTM is reporting "check failed" because "No code changes detected". Is this expected? Can we avoid somehow this misleading message?
> Screen Shot 2019-11-01 at 11 23 34 PM

Hm, this is new, originally the message said "X out of Y checks passed", so this should be definitely fixed on GitHub's side (the LGTM check is in a "neutral" state, so it didn't pass nor failed).

mrc0mmand:
@anakryiko also, one last thing - you have to setup a cron job in Travis CI to make Coverity work (if you haven't already done that):

image

anakryiko:
@mrc0mmand, done

PR #94, opened 2019-11-05
coverity: explicitly use bash instead of sh

mrc0mmand:
On Ubuntu `/bin/sh` is a symlink to `/bin/dash`, which doesn't support
certain builtins used by the Coverity script (namely pushd/popd)

---

Hopefully a last follow-up for the Coverity-related stuff.

### DIFF
--- a/scripts/coverity.sh
+++ b/scripts/coverity.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # Taken from: https://scan.coverity.com/scripts/travisci_build_coverity_scan.sh
 # Local changes are annotated with "#[local]"
 


mrc0mmand:
Judging by the latest Travis CI [cron job](https://travis-ci.org/libbpf/libbpf/jobs/608391729) and a notification email from Coverity about new defects I'd say the Coverity integration finally works as expected!

evverx:
I've just received that notification about a memory leak as well so it seems to be working. Thank you!

What I don't understand though is why it was found by Coverity Scan today. Judging by `git log` that memory leak has been there for at least 10 months. My first guess would be that I slipped up building `libbpf` somewhere when I first sent it there :-)

yonghong-song:
There is no memory leak here. `prog->insns = new_insn` assigned the new_insn (new memory) to `prog->insns`. insns will be freed later.

evverx:
@yonghong-song that's good to know. Thank you!

What I was trying to say is that (even if it's a false positive) it's weird that report has just surfaced.

evverx:
@yonghong-song I took a closer look at the report and it appears `new_insn` is assigned to `prog->insns` after that "return" Coverity is complaining about and I can't seem to find any implicit "destructors" freeing `new_insn`. I'm wondering if we're talking about the same Coverity report. Anyway, I'm not sure where `new_insn` is freed in the following code (though, it could be I'm just missing something):
```c
		new_insn = reallocarray(prog->insns, new_cnt, sizeof(*insn));
		if (!new_insn) {
			pr_warn("oom in prog realloc\n");
			return -ENOMEM;
		}

		if (obj->btf_ext) {
			err = bpf_program_reloc_btf_ext(prog, obj,
							text->section_name,
							prog->insns_cnt);
			if (err)
				return err;
		}
```

yonghong-song:
Oh, you are right. I mean `prog->insns` is freed eventually. But in the above error path inside `obj->btf_ext`, we do have a problem for memory leak and double free since later on when prog->insns to be freed, it will be a double free.

anakryiko:
Fixed this and a bunch of other found issues in https://patchwork.ozlabs.org/project/netdev/list/?series=141246&state=*

evverx:
@anakryiko thank you!

Meanwhile, I've been trying to figure out why I can't reproduce that report using the latest coverity binary and my local instance. No luck so far :-) On the bright side, it appears whatever is going on on Travis CI here is correct and it would probably make sense to make sure `systemd` doesn't miss this kind of reports (though as far as I can tell those parts of `coverity.sh` are almost identical). @mrc0mmand thank you again!

PR #99, opened 2019-11-21
include <linux/stddef.h> to get macro defintions

dheater:
include <linux/stddef.h> to get macro defintions for compiling on Alpine Linux 3.10

Otherwise, macros like `__always_inline` and `__cpu_to_be64p` are undefined.

### DIFF
--- a/src/libbpf.c
+++ b/src/libbpf.c
@@ -24,6 +24,7 @@
 #include <fcntl.h>
 #include <errno.h>
 #include <asm/unistd.h>
+#include <linux/stddef.h> // Not alphabetical - Get required macro definitions
 #include <linux/err.h>
 #include <linux/kernel.h>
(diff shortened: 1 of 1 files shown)


anakryiko:
Libbpf itself doesn't use `__always_inline` or `__cpu_to_be64p`. Where exactly you get compiler error for these?

Also, this change will have to go through kernel repo as a patch to bpf@vger.kernel.org.

dheater:
Ah. Interesting. I see these errors when I try to build `bcc` which includes `libbpf` as a submodule.
But if I build `libbpf` from the same Docker container, it builds fine.

Looks like the problem is with how `bcc` builds.

dheater:
Looks like this was the same issue as https://github.com/libbpf/libbpf/pull/41

PR #102, opened 2019-12-01
Add -fPIC for static objects

olsajiri:
When using the libbpf.a in bcc's shared library build
I'm getting following compile error:

  [ 10%] Linking CXX shared library libbcc.so
  /usr/bin/ld: libbcc_bpf.a(libbpf.o): relocation R_X86_64_PC32 against       \
    symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; \
    recompile with -fPIC

Adding the -fPIC also for static objects, so it can
be used to build position independent code.

Signed-off-by: Jiri Olsa <jolsa@kernel.org>

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -16,6 +16,7 @@ ifneq ($(FEATURE_REALLOCARRAY),)
 endif
 
 SHARED_CFLAGS += -fPIC -fvisibility=hidden -DSHARED
+STATIC_CFLAGS += -fPIC
 
 CFLAGS ?= -g -O2 -Werror -Wall
 ALL_CFLAGS += $(CFLAGS) -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

anakryiko:
@olsajiri, I think it's bad to hard-code -fPIC build for libbpf.a, this can decrease code performance for user applications linking statically against libbpf (and not caring about being position-independent). You should be able to fix your build problems with overriding CFLAGS when building libbpf (e.g., `CFLAGS=-g -O2 -Werror -Wall -fPIC`).

olsajiri:
hum right, I don't like the idea of switching it on for everybody,
but I need that for the libbcc.so to compile and I think there might
be others in future.. also this seems to be an issue for x86_64 only

how about we package 2 static libbpf.a versions with and without -fPIC?
but I guess on the Fedora rpm spec level anyway.. ;-)

anakryiko:
I think we should keep Makefile as simple as possible w/o extra options/flags/etc, if we can avoid them. At Facebook we also have a need for both PIC and non-PIC static libraries, but we are easily accomplishing this by using current version of Makefile. So I'm suspecting it's not hard to do that for your use case as well (especially that libbcc doesn't and shouldn't use any non-exported APIs).

olsajiri:
yep, thanks

PR #106, opened 2019-12-11
Makefile: Add cscope and tags rules

tohojo:
These were added to the kernel repo, but not in Github. However, they are
useful for browsing the source in Github while prototyping new features and
compiling them into userspace utilities.

Signed-off-by: Toke Høiland-Jørgensen <toke@redhat.com>


anakryiko:
thanks!

PR #126, opened 2020-02-21
ci: clean up .travis.yaml, build latest pahole

anakryiko:
Clean up Travis CI config, extract multi-step initializations into scripts.
Also, move kernel-building tests to happen last to not block lightweight
Debian and Ubuntu tests.

This PR also adds step to build pahole from latest sources. It also switches to LLVM 11. It also drops 5.5.0-rc6, which was kept only because it was built with pahole 1.16 and had all the latest BTF features. Now LATEST kernel has that, so there is no need for 5.5.0-rc6 anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ addons:
             - elfutils
             - libcap-dev
             - libelf-dev
+            - libdw-dev
 
 stages:
     # Run Coverity periodically instead of for each PR for following reasons:
@@ -94,13 +95,6 @@ jobs:
           name: Kernel 5.5.0 + selftests
           language: bash
           env: KERNEL=5.5.0
-          script:  $CI_ROOT/vmtest/run_vmtest.sh || travis_terminate
-
-        # 5.5.0-rc6 is built from bpf-next; TODO(hex@): remove when pahole v1.16 is available
-        - name: Kernel 5.5.0-rc6 (pahole 1.16) + selftests
-          language: bash
-          env: KERNEL=5.5.0-rc6
-          install: sudo adduser "${USER}" kvm
           script:  $CI_ROOT/vmtest/run_vmtest.sh || travis_terminate
 
         - name: Kernel LATEST + selftests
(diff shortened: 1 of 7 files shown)


travis-ci/vmtest/run_vmtest.sh
```diff
@@ -0,0 +1,27 @@
...
+
+echo "KERNEL: $KERNEL"
+
+# Install required packages
+wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main" | sudo tee -a /etc/apt/sources.list
+echo "deb http://archive.ubuntu.com/ubuntu eoan main restricted universe multiverse" | sudo tee -a /etc/apt/sources.list
+sudo apt-get -qq update
+sudo apt-get -y install dwarves=1.15-1
+sudo apt-get -qq -y install clang-10 lld-10 llvm-10
```
wat-ze-hex: Thinking about further improvements, not related to this PR:
Since we're looking into building pahole from the sources it would be great to have it built once and reused for each kernel. Same for other steps which don't depend on kernel. This would save plenty of time.

anakryiko: Building pahole shouldn't take long. It's a simple and small binary, so I'm not sure it's worth doing this.

.travis.yml
```diff
@@ -6,18 +6,11 @@ services:
...
-        - VMTEST_SETUPCMD="PROJECT_NAME=${PROJECT_NAME} ./${PROJECT_NAME}/travis-ci/vmtest/run_selftests.sh"
-    jobs:
-        # Setup command override.
-        # 5.5.0-rc6 is built from bpf-next; TODO(hex@): remove when pahole v1.16 is available
-        - KERNEL=5.5.0-rc6
-        - KERNEL=5.5.0
-        - KERNEL=LATEST
+        - VMTEST_ROOT="$REPO_ROOT/travis-ci/vmtest"
+        - PROJECT_NAME='libbpf'
+        - SCRIPTS="$TRAVIS_BUILD_DIR/travis-ci"
```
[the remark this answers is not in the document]
anakryiko: ok, will update

PR #137, opened 2020-03-10
Update disk image

wat-ze-hex:
- Update disk image
- Update blacklist
- Run `test_maps` and `test_verifier` in addition to `test_progs`
- Add `mkrootfs.sh` image build script

### DIFF
--- a/travis-ci/vmtest/mkrootfs.sh
+++ b/travis-ci/vmtest/mkrootfs.sh
@@ -1,0 +1,146 @@
+#!/bin/bash
+
+# This script is based on drgn script for generating Arch Linux bootstrap
+# images.
+# https://github.com/osandov/drgn/blob/master/scripts/vmtest/mkrootfs.sh
+
+set -euo pipefail
+
+usage () {
+	USAGE_STRING="usage: $0 [NAME]
+       $0 -h
+
+Build an Arch Linux root filesystem image for testing libbpf in a virtual
+machine.
+
+The image is generated as a zstd-compressed tarball.
+
+This must be run as root, as most of the installation is done in a chroot.
+
+Arguments:
+  NAME   name of generated image file (default:
+         libbpf-vmtest-rootfs-\$DATE.tar.zst)
+
+Options:
+  -h     display this help message and exit"
+
+	case "$1" in
+		out)
+			echo "$USAGE_STRING"
+			exit 0
+			;;
+		err)
+			echo "$USAGE_STRING" >&2
+			exit 1
+			;;
+	esac
+}
+
+while getopts "h" OPT; do
+	case "$OPT" in
+		h)
+			usage out
+			;;
+		*)
+			usage err
+			;;
+	esac
+done
+if [[ $OPTIND -eq $# ]]; then
+	NAME="${!OPTIND}"
+elif [[ $OPTIND -gt $# ]]; then
+	NAME="libbpf-vmtest-rootfs-$(date +%Y.%m.%d).tar.zst"
+else
+	usage err
+fi
+
+pacman_conf=
+root=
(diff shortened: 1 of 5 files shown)


1
```diff
@@ -0,0 +1,18 @@
+vmtest: update disk image
```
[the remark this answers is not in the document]
wat-ze-hex: > hmm.. what is this file? github UI doesn't show the name :)

That was `git add .` Thanks for noticing.

travis-ci/vmtest/run_selftests.sh
```diff
@@ -1,11 +1,22 @@
...
+test_maps() {
+	echo TEST_MAPS
+	# Allow failing on older kernels.
+	./test_maps || [ ${KERNEL} != 'LATEST' ]
+}
+
+test_verifier() {
+	echo TEST_VERIFIER
+	./test_verifier
+}
```
anakryiko: let's not just run test_maps and test_verifier not on latest. They have no built-in mechanisms to blacklist latest tests, so we won't be able to rely on them for older kernel anyway. So why bother and prolong tests.

travis-ci/vmtest/mkrootfs.sh
```diff
@@ -0,0 +1,146 @@
...
+EOF
+chmod 644 "$root/etc/inittab"
+
+mkdir -m 755 "$root/etc/init.d" "$root/etc/rcS.d"
+cat > "$root/etc/rcS.d/S40network" << "EOF"
+#!/bin/sh
+
+ip link set lo up
+EOF
+chmod 755 "$root/etc/rcS.d/S40network"
```
anakryiko: nit: run.sh uses "S40-network" naming convention

travis-ci/vmtest/mkrootfs.sh
```diff
@@ -0,0 +1,146 @@
...
+# We don't need D, Fortran, or Go.
+ rm -f "$root/usr/lib/libgdruntime."* \
+	"$root/usr/lib/libgphobos."* \
+	"$root/usr/lib/libgfortran."* \
+	"$root/usr/lib/libgo."*
+# We don't need any documentation.
+rm -rf "$root/usr/share/doc" \
+	"$root/usr/share/help" \
+	"$root/usr/share/man" \
+	"$root/usr/share/texinfo"
```
anakryiko: rm -rf "$root/usr/share/{doc,help,man,texinfo}" should work as well

travis-ci/vmtest/run_selftests.sh
```diff
@@ -20,5 +31,13 @@ fi
...
-echo TEST_PROGS
-./test_progs ${BLACKLIST:+-b$BLACKLIST} ${WHITELIST:+-t$WHITELIST}
+set +e
+exitcode=0
+for test_func in test_progs test_maps test_verifier; do
+	${test_func}; c=$?
+	if [[ $c -ne 0 ]]; then
+		exitcode=$c
+	fi
+done
```
anakryiko: there is `set -e` on line 3, not sure why you need this instead of just:

```
test_progs
test_maps
test_verifier
```

wat-ze-hex: This is done to run all of `test_progs`, `test_maps`, `test_verifier` even if any of them fails. That's why `set +e` is set.

anakryiko: Ah, I see. I think it's going to be painful in practice, though, because if test_progs fails, but subsequent test_maps and test_verifiers passes, it will be a pain to find the failure. Let's keep -e and run them one by one until first failure?

wat-ze-hex: ok, will change in a follow-up.

PR #145, opened 2020-04-03
Install iproute2

joestringer:
This allows certain tests like the sk_assign selftests to run.

I tried out these commands via a docker container something like the following, but I don't know how to actually use the mkrootfs here without some arch-like environment (it fails in the docker container run like below):

```
$ docker run --tty --interactive --rm archlinux bash
```

Related: #144

PR #144, opened 2020-04-02
ci: update image index and enable sk_assign selftest

anakryiko:
Image was updated to include iproute2 package, which is a requirement for
sk_assign selftests. So enable it now.

Signed-off-by: Andrii Nakryiko <andriin@fb.com>

PR #147, opened 2020-04-11
vmtest: update root fs, whitelist sk_assign test

wat-ze-hex:
1. Update mkrootfs.sh building root fs
- Remove /etc/fstab from root fs and mount each fs type separately in
S10-mount script.
- devtmpfs can be already mounted prior to S10-mount execution so make
it opt-out. This addresses [0].
- set -eux for scripts
2. Add iproute2 to root fs and whitelist sk_assign test. Addresses
[1][2]. Update INDEX file with 2020-09-27 version.

[0] https://github.com/libbpf/libbpf/pull/145#issuecomment-609673493
[1] https://github.com/libbpf/libbpf/pull/144
[2] https://github.com/libbpf/libbpf/pull/145

### DIFF
--- a/travis-ci/vmtest/configs/blacklist/BLACKLIST-latest
+++ b/travis-ci/vmtest/configs/blacklist/BLACKLIST-latest
@@ -1,7 +1,5 @@
 # TEMPORARILY DISABLED
 send_signal		# flaky
 test_lsm		# semi-working
-sk_assign		# needs better setup in Travis CI
-sk_lookup
 core_reloc		# temporary test breakage
 bpf_verif_scale		# clang regression
--- a/travis-ci/vmtest/mkrootfs.sh
+++ b/travis-ci/vmtest/mkrootfs.sh
@@ -81,6 +81,7 @@ packages=(
 	binutils
 	elfutils
 	glibc
+	iproute2
 	# selftests test_verifier dependencies.
 	libcap
 )
@@ -101,15 +102,6 @@ rm -rf "$root/usr/share/{doc,help,man,texinfo}"
 
 chroot "${root}" /bin/busybox --install
 
-cat > "$root/etc/fstab" << "EOF"
-dev /dev devtmpfs rw,nosuid 0 0
-proc /proc proc rw,nosuid,nodev,noexec 0 0
-sys /sys sysfs rw,nosuid,nodev,noexec 0 0
-debugfs /sys/kernel/debug debugfs mode=755,realtime 0 0
-bpffs /sys/fs/bpf bpf realtime 0 0
-EOF
-chmod 644 "$root/etc/fstab"
-
 cat > "$root/etc/inittab" << "EOF"
 ::sysinit:/etc/init.d/rcS
 ::ctrlaltdel:/sbin/reboot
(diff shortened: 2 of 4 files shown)


travis-ci/vmtest/mkrootfs.sh
```diff
@@ -123,20 +115,34 @@ mkdir -m 755 "$root/etc/init.d" "$root/etc/rcS.d"
 cat > "$root/etc/rcS.d/S10-mount" << "EOF"
 #!/bin/sh
 
-/bin/mount -a
+set -eux
+
+/bin/mount proc /proc -t proc
+# Try mounting devtmpfs
+/bin/mount devtmpfs /dev -t devtmpfs || true
```
anakryiko, line 122: why `|| true` here? is it expected to fail?

travis-ci/vmtest/mkrootfs.sh
```diff
@@ -123,20 +115,34 @@ mkdir -m 755 "$root/etc/init.d" "$root/etc/rcS.d"
...
+/bin/mount sysfs /sys -t sysfs
+/bin/mount bpffs /sys/fs/bpf -t bpf
+/bin/mount debugfs /sys/kernel/debug -t debugfs
+
+echo 'Listing currently mounted file systems'
+/bin/mount
```
anakryiko, line 128: debug leftover or you'd like to leave that here permanently?

travis-ci/vmtest/configs/blacklist/BLACKLIST-latest
```diff
@@ -1,7 +1,6 @@
 # TEMPORARILY DISABLED
 send_signal		# flaky
 test_lsm		# semi-working
-sk_assign		# needs better setup in Travis CI
 sk_lookup
```
anakryiko, line 4: can you try enabling sk_lookup as well? I'm not sure, but it might have been disabled due to the same iproute2 reason.

anakryiko (commented):
Thanks for the update, seems like we might finally get iproute2-dependent tests running, yay! Please check if sk_lookup also works.

travis-ci/vmtest/mkrootfs.sh
wat-ze-hex: To address the case when `devtmpfs` is mounted prior to `S10-mount` run:
```
+ '[' -x /etc/rcS.d/S10-mount ']'
+ /etc/rcS.d/S10-mount
+ /bin/mount proc /proc -t proc
+ /bin/mount devtmpfs /dev -t devtmpfs
mount: mounting devtmpfs on /dev failed: Resource busy
```

travis-ci/vmtest/mkrootfs.sh
wat-ze-hex: I think it's useful to have mounts listed in CI execution log, since I've run into mounting issues while bumping this version of disk image. Wdyt?

travis-ci/vmtest/mkrootfs.sh
wat-ze-hex: I'll do a check before mounting instead of silencing with `|| true`.

travis-ci/vmtest/mkrootfs.sh
anakryiko: yeah, that's fine, especially that now logs are collapsed anyway

anakryiko:
Awesome, thanks!

PR #156, opened 2020-06-03
README: add BTF and CO-RE info

anakryiko:
Add list of Linux distributions with kernel BTF built-in.
Give few useful links to BPF CO-RE-related material to help users get started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ $ PKG_CONFIG_PATH=/build/root/lib64/pkgconfig DESTDIR=/build/root make install
 ```
 
 Distributions
-=====
+=============
 
 Distributions packaging libbpf from this mirror:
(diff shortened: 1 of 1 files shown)


anakryiko:
@borkmann, @netoptimizer, @tohojo FYI

anakryiko:
@brendangregg  as well

tohojo:
Thanks for the heads up! A few comments:

- RHEL 8.2 only has LLVM9
- Ubuntu has LLVM10 from v20.04, and Arch Linux also packages LLVM10.

Opened a bug report for the Arch kernel config here: https://bugs.archlinux.org/task/66877

anakryiko:
My bad about RHEL 8.2 and LLVM, I actually got this info from Tom Stellard, but somehow screw up in README. I'll add Ubuntu to the list for LLVM.

PR #174, opened 2020-07-01
travis_ci: run tests for 4.9 kernel

wat-ze-hex:
Make sure that libbpf sanitizes BTF properly for older kernels.
Add a stage for 4.9.0 kernel in TravisCI.
For now make test failures non-blocking by adding 4.9.0 to `allow_failures`
section.
Blacklist is copy-pasted 5.5.0 kernel blacklist.


anakryiko:
Thanks!

PR #181, opened 2020-07-27
check-reallocarray.sh: Use the same compiler Make does

tchebb:
Currently we hardcode "gcc", which means we get a bogus result any time a non-default CC is passed to Make. In fact, it's bogus even when `CC` is not explicitly set, since Make's default is "cc", which isn't necessarily the same as "gcc".

Fix the issue by passing the compiler to use to `check-reallocarray.sh`.


anakryiko:
LGTM. Thanks.

PR #201, opened 2020-09-07
remove internal reallocarray()

vcunat:
~For reference, we ran into this problem due to strict sandboxing in [a packaging solution](https://github.com/nixos/nixpkgs).~

... as it's covered by libbpf_reallocarray() since commit dc70da9c70.


anakryiko:
libbpf doesn't require check-reallocarray.sh anymore, would you mind instead just removing it?

vcunat:
I certainly wouldn't mind.   I ran into 0.1.0 failing due to `/tmp` being blocked by sandbox and I actually haven't even checked against master (given the freshness of the release).

anakryiko:
@vcunat, do you plan to do a follow up and remove check-reallocarray.sh altogether?

vcunat:
Ah, I thought you/someone planned to do it, but now I see that it wasn't clear from your post.  So, I did not plan it, but I think I could do it within a couple days.

vcunat:
I expect you wanted something like this?

src/libbpf_internal.h
```diff
@@ -83,11 +83,8 @@ do {				\
 #endif
 /*
  * Re-implement glibc's reallocarray() for libbpf internal-only use.
- * reallocarray(), unfortunately, is not available in all versions of glibc,
- * so requires extra feature detection and using reallocarray() stub from
- * <tools/libc_compat.h> and COMPAT_NEED_REALLOCARRAY. All this complicates
- * build of libbpf unnecessarily and is just a maintenance burden. Instead,
- * it's trivial to implement libbpf-specific internal version and use it
```
anakryiko, line 90: please don't change this file, it gets auto-synced

scripts/sync-kernel.sh
```diff
@@ -45,7 +45,6 @@ PATH_MAP=(									\
 	[tools/include/uapi/linux/if_link.h]=include/uapi/linux/if_link.h	\
 	[tools/include/uapi/linux/if_xdp.h]=include/uapi/linux/if_xdp.h		\
 	[tools/include/uapi/linux/netlink.h]=include/uapi/linux/netlink.h	\
-	[tools/include/tools/libc_compat.h]=include/tools/libc_compat.h		\
```
anakryiko, line 48: can you please remove this file from include/tools as well?

anakryiko (commented):
almost :)

scripts/sync-kernel.sh
vcunat: I did remove `include/tools/libc_compat.h`... so you missed that or I don't understand?

src/libbpf_internal.h
vcunat: Ah, I wasn't watching out for that.  Amended.

scripts/sync-kernel.sh
anakryiko: ah, my bad, missed it in "Files changed" view, sorry about that

vcunat: No problem :-)

PR #202, opened 2020-09-08
Makefile: link against zlib

vcunat:
Without this we would be missing symbols, as shown e.g. by
ldd -r libbpf.so

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -24,8 +24,8 @@ ifdef NO_PKG_CONFIG
 	ALL_LDFLAGS += -lelf -lz
 else
 	PKG_CONFIG ?= pkg-config
-	ALL_CFLAGS += $(shell $(PKG_CONFIG) --cflags libelf)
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -24,8 +24,8 @@ ifdef NO_PKG_CONFIG
 	ALL_LDFLAGS += -lelf -lz
 else
 	PKG_CONFIG ?= pkg-config
-	ALL_CFLAGS += $(shell $(PKG_CONFIG) --cflags libelf)
-	ALL_LDFLAGS += $(shell $(PKG_CONFIG) --libs libelf)
+	ALL_CFLAGS += $(shell $(PKG_CONFIG) --cflags libelf --cflags zlib)
```
anakryiko, line 27: would just `--cflags libelf zlib` work? no need to repeat --cflags flag?

vcunat: Right.  I tested it (with 0.1.0) and also `man pkg-config` suggests so.  I agree it will be better; updated.

PR #204, opened 2020-09-14
travis-ci: add s390x to CI

iii-i:
Hi,

I've been able to start `test_verifier` locally with these patches. It currently dies with a kernel panic, but it's a start.

```
------------[ cut here ]------------
trace type BPF program uses run-time allocation
illegal operation: 0001 ilc:2 [#1] SMP 
Modules linked in:
CPU: 0 PID: 62 Comm: test_verifier Not tainted 5.9.0-rc1 #1
Hardware name: QEMU 2964 QEMU (KVM/Linux)
Krnl PSW : 0704d00180000000 00000000002ec512 (replace_map_fd_with_map_ptr+0x58a/0x690)
           R:0 T:1 IO:1 EX:1 Key:0 M:1 W:0 P:0 AS:3 CC:1 PM:0 RI:0 EA:3
Krnl GPRS: 00000000ffff7cf6 0000000000cf88d0 000000000000002f 0000000000000006
           0000000000000000 000003e000227930 000003ff00000005 000003e000000004
           000000007b5a4000 000003e00003d000 000000007b079000 000003e00003d058
           000000007b050000 000000007b079000 00000000002ec50e 000003e000227b00
Krnl Code: 00000000002ec502: c02000443fc6	larl	%r2,0000000000b7448e
           00000000002ec508: c0e5fff48b80	brasl	%r14,000000000017dc08
          #00000000002ec50e: af000000		mc	0,0
          >00000000002ec512: a7f4fe1e		brc	15,00000000002ec14e
           00000000002ec516: b916004d		llgfr	%r4,%r13
           00000000002ec51a: b9040028		lgr	%r2,%r8
           00000000002ec51e: c0300044407e	larl	%r3,0000000000b7461a
           00000000002ec524: c0e5fffffcee	brasl	%r14,00000000002ebf00
Call Trace:
 [<00000000002ec512>] replace_map_fd_with_map_ptr+0x58a/0x690 
([<00000000002ec50e>] replace_map_fd_with_map_ptr+0x586/0x690)
 [<00000000002fae18>] bpf_check+0x230/0x17c0 
 [<00000000002e66be>] bpf_prog_load+0x45e/0x888 
 [<00000000002e7992>] __do_sys_bpf+0x742/0x17f0 
 [<00000000009bac74>] system_call+0xdc/0x2b8 
INFO: lockdep is turned off.
Last Breaking-Event-Address:
 [<000000000017dc68>] __warn_printk+0x60/0x68
Kernel panic - not syncing: Fatal exception: panic_on_oops
```

I guess this has something to do with QEMU and `MONITOR CALL` instruction. I'll look at it in more detail later.

I also didn't have a chance to run this in an actual Travis environment. So let's see how it goes.


travis-ci/vmtest/run_selftests.sh
```diff
@@ -43,9 +43,9 @@ fi
 
 cd libbpf/selftests/bpf
 
-test_progs
+[[ "$(uname -m)" != s390x ]] && test_progs
```
iii-i, line 46: Hm, this doesn't look right: should be `[[ "$(uname -m)" == s390x ]] || test_progs`

iii-i:
@anakryiko Regarding the current x86_64 failure: it's most likely related to Ubuntu 20.04 upgrade. Due to CET `bpf_fentry_test1` looks like this:

```
   0xffffffff816929a0 <bpf_fentry_test1>:	endbr64 
   0xffffffff816929a4 <bpf_fentry_test1+4>:	nopl   0x0(%rax,%rax,1)
```

and as a result `__bpf_arch_text_poke` doesn't like it. I guess this needs to be fixed in the kernel?

anakryiko:
I didn't have a chance to look at this thoroughly yet. But few questions from a cursory look:

1. Why did you have to update Ubuntu to focal?
2. Why do you need to build a bunch of stuff that we don't build from sources for x86_64?
3. All those s390x cross-compilation packages you added to .travis.yaml. Is it possible to contain them to just s390x target?
4. Assuming we absolutely have to do extra source code builds, we probably will need to figure out how to structure build_selftests.sh script better so that arch-specific parts are in separate scripts. Right now s390x special-handling is overwhelming there and obscures the flow.

iii-i:
1. bionic qemu-system-s390x is too old, it does not support s390 model that the modern kernels require.
2. Because they are libbpf/selftest dependencies, and don't seem to be installable using multiarch in Ubuntu. Btw, in Debian I can install them just fine.
3. It doesn't seem possible with apt addon, but I can always call apt manually.
4. I can move this stuff out, it's no big deal.

CET/-fcf-protection is in general an interesting topic, because I don't see ftrace support for it either.

Regarding s390x `mc` failure, it's indeed a missing instruction in qemu and unfortunately it might take a while to get this fixed. And even then I don't know whether this would be backported, so in the worst case we would have to build qemu as well.

iii-i:
- Added `build_qemu` in order to build qemu-system-s390x that is compatible with latest kernels.
- Pulled focal upgrade - since we build qemu anyway, we don't need it. Still, CET support in ftrace and bpf is an interesting topic..
- Moved cross-compiler installation to s390x stage.
- Moved cross-compilation stuff to `build_deps.sh`.

iii-i:
- Use a different elfutils URL (should hopefully fix `Failed to connect to sourceware.org port 21: No route to host`).

Btw, there are a few failures in `test_verifier` because the kernel has no BTF.
I guess we need to skip those:
```
#452/p d_path accept FAIL
#453/p d_path reject FAIL
#705/p bpf_map_ptr: read with negative offset rejected FAIL
#706/p bpf_map_ptr: write rejected FAIL
#707/p bpf_map_ptr: read non-existent field rejected FAIL
#708/p bpf_map_ptr: read ops field accepted FAIL
#712/p calls: two calls returning different map pointers for lookup (hash, map in map) FAIL
```

iii-i:
- Build libcap with `PAM_CAP=no` to avoid cross-compiling libpam (libcap detects whether or not to build PAM module by checking `/usr/include/security/pam_modules.h` availability).
- Posted https://lore.kernel.org/bpf/20200921184219.4168733-1-iii@linux.ibm.com/ - with this patch `test_verifier` passes.

iii-i:
- Fixed qemu checkout sequence.

anakryiko:
Ilya, I think you are approaching this wrong with over-complicated cross-arch approach. Travis CI supports s390x architecture natively, why don't you use it to build libbpf, selftests and kernel and then running s390x-native qemu with s390x latest kernel? See "Ubuntu Bionic Build (s390x)" configuration, notice `arch: s390x` part.

The big downside to compiling all the dependencies from scratch is unreliability and a very long times to do that. Notice that we don't compile clang from sources, that would take way too long. I also stripped down kernel configuration significantly to reduce build times. The only thing that we build from scratch is pahole, because often we need the bleeding-edge feature that hasn't been released yet. But otherwise, I'd rather keep it fast and simple with pre-packaged libraries. So please try to do everything natively on s390x platform. That should eliminate a lot of your pain points.

iii-i:
Oh, I assumed Travis was using qemu tcg and therefore the kernel build would be unreasonably slow. But if it's native, this changes everything. I'll give it at try.

iii-i:
- Use native s390x builder.
- Use `$(uname -m)` instead of `$LIBBPF_ARCH`.
- Enable `test_maps`.

iii-i:
```
E: Unable to locate package clang-12
E: Unable to locate package lld-12
E: Unable to locate package llvm-12
```

Seems like LLVM does not publish s390x packages. I'll ask our LLVM folks whether this could be improved.

anakryiko:
@iii-i do you still plan to work on this or should be close this PR?

iii-i:
Hi! Sorry about the big delay, this is still on my list. The current problem is that we don't have continuously updated IBM Z LLVM packages for the `main` branch in the official apt repo. Unfortunately the maintainers don't have the time to work on such continuous build or to maintain it, so this is something I'm supposed to do. Hopefully I'll find the time to do this early this year..

iii-i:
A short update:
* The discussion with the apt.llvm.org maintainer has been revived, and there is now a chance that s390x packages will be published there at some point. As an interim solution I made the https://launchpad.net/~iii-i/+archive/ubuntu/llvm-toolchain-systemz PPA, which is built using the same scripts as apt.llvm.org.
* The BPF testcases currently fail to build due to https://reviews.llvm.org/D83289, we need to resolve this first.
* There are a few failures in `test_verifier` due to atomic insns, I'll need to resolve this first too. After that we can build everything and run test_verifier.

anakryiko:
> A short update:
> 
>     * The discussion with the apt.llvm.org maintainer has been revived, and there is now a chance that s390x packages will be published there at some point. As an interim solution I made the https://launchpad.net/~iii-i/+archive/ubuntu/llvm-toolchain-systemz PPA, which is built using the same scripts as apt.llvm.org.

ok, sounds good

> 
>     * The BPF testcases currently fail to build due to https://reviews.llvm.org/D83289, we need to resolve this first.

left few comments on that diff

> 
>     * There are a few failures in `test_verifier` due to atomic insns, I'll need to resolve this first too. After that we can build everything and run test_verifier.

as in runtime failures or compilation failures? But regardless, yes, atomics is a new thing that is platform-specific. BTW, what about fentry/fexit programs, relying on BPF trampoline. Does that work on s390x?

iii-i:
Runtime failures. JIT doesn't know how to handle new opcodes, plus I think there is a common code problem with zero-extending 32-bit BPF_FETCH results. I'm currently preparing the patch series for all that.

Trampolines don't work yet, among other things because our vmlinux (unlike amd64) is too far away from modules, so we can't easily jump between it and JITed code. Our "core" kernel guys are currently working on lifting that restriction.

iii-i:
Rebased and cleaned up the patches a little. Locally this worked fine.

Regarding the "vmtests: add s390x image" patch: you might want to rebuild the image yourself (should be as simple as running `mkrootfs_debian.sh` - you should still have access to the `ebpf01` s390x machine) and upload the result to your S3 bucket - I can then adjust the patch accordingly.

anakryiko:
> Rebased and cleaned up the patches a little. Locally this worked fine.
> 
> Regarding the "vmtests: add s390x image" patch: you might want to rebuild the image yourself (should be as simple as running `mkrootfs_debian.sh` - you should still have access to the `ebpf01` s390x machine) and upload the result to your S3 bucket - I can then adjust the patch accordingly.

Awesome. I'll try to get to this in next day or two, I've accumulated quite a backlog :( Stay tuned!

iii-i:
```
Preparing Linux image

Root filesystem version: 2021.03.24

Disk image: /home/travis/root.img

mount: /home/travis/mnt.6j4k8WUsv7: mount failed: Operation not permitted.
```

Hmm, this is really strange. `mount` runs with `sudo`, so we are root. I saw a similar problem locally when testing these scripts in docker, fixed by doing `--privileged` + `-v /dev:/dev`. But here we are surely running inside our own VM?

travis-ci/vmtest/run_selftests.sh
```diff
@@ -43,7 +43,7 @@ fi
 
 cd libbpf/selftests/bpf
 
-test_progs
+[[ "$(uname -m)" == s390x ]] || test_progs
```
anakryiko, line 46: let's do explicit if and leave a comment why we don't run test_progs on s390x. And yeah, why? I thought the whole point was to run them

travis-ci/vmtest/run.sh
```diff
@@ -243,7 +243,7 @@ create_rootfs_img() {
 download_rootfs() {
 	local rootfsversion="$1"
 	local dir="$2"
-	download "${PROJECT_NAME}-vmtest-rootfs-$rootfsversion.tar.zst" |
+	download "$(uname -m)/${PROJECT_NAME}-vmtest-rootfs-$rootfsversion.tar.zst" |
```
anakryiko, line 246: instead of inlining `uname -m`, let's extract it into ARCH variable and use it consistently throughout all scripts?

anakryiko (commented):
some very brief comments while briefly skimming through the changes

travis-ci/vmtest/run_selftests.sh
iii-i: Currently the trampoline implementation is still missing, but I think having the infrastructure and running test_verifier is nevertheless a good first step. I will add a comment here.

travis-ci/vmtest/run.sh
iii-i: OK.

travis-ci/vmtest/run_selftests.sh
anakryiko: see how we do blacklisting/whitelisting for old kernels. We can have black-/whitelists per architecture, if necessary, so you'll be able to enable most tests expect few that you know should not work.

But you are signing up to maintaining this, of course ;) It's the most unpleasant part of libbpf syncs: updating blacklists.

travis-ci/vmtest/run.sh
iii-i: What do you think about putting `ARCH=$(uname -m)` into `vmtest/helpers.sh`? An alternative option would be to require this variable to be set externally, but this would complicate running individual scripts.

travis-ci/vmtest/run_selftests.sh
iii-i: Sounds good. I think I'll add a couple `$ARCH` subdirectories under `config/` for that.

travis-ci/vmtest/run.sh
anakryiko: Yeah, helpers.sh is supposed to have a shared functionality between all scripts, so this makes sense.

travis-ci/vmtest/run_selftests.sh
anakryiko: so if you are going to rework that part, I was always a bit annoyed by a deeply-nested structure of configs. Instead of `configs/blacklist/BLACKLIST-5.5.0` I'd rather have just `BLACKLIST-5.5.0` and `WHITELIST-4.9.0` without directory nesting. So with arch we probably want something like `BLACKLIST-latest.s390x` with fallback (if arch-specific is missing, I don't think we need to merge) to `BLACKLIST-latest` that is common for all arches. WDYT?

travis-ci/vmtest/run.sh
anakryiko: common.sh might be better name, though

travis-ci/vmtest/run_selftests.sh
iii-i: Ah, I actually did that for configs initially, but then I thought that having directories is better for grouping (e.g. you won't have to look at s390x stuff when working with x86_64), so in my working copy I now have this:

```
├── INDEX
├── s390x
│   ├── blacklist
│   │   └── BLACKLIST-latest
│   └── latest.config
└── x86_64
    ├── blacklist
    │   ├── BLACKLIST-5.5.0
    │   └── BLACKLIST-latest
    ├── latest.config
    └── whitelist
        └── WHITELIST-4.9.0
```

But on the other hand removing blacklist/whitelist nesting makes sense. So what do you think about something like the following?

```
├── INDEX
├── s390x
│   ├── BLACKLIST-latest
│   └── latest.config
└── x86_64
    ├── BLACKLIST-5.5.0
    ├── BLACKLIST-latest
    ├── latest.config
    └── WHITELIST-4.9.0
```

anakryiko: I think it's not a lot of configs to need separate per-arch directory, honestly. And I take back what I said about not needing to merge common and per-arch config. I think both are useful because sometimes I have to disable known-bad selftest until it is fixed upstream (and merged back into bpf-next). So having a common place where I need to add one entry once is good. But we don't have to really merge them, it's easier to add support for specifying multiple whitelists and blacklists to test_progs.

So overall, I still think .arch suffix and flat configs directory is better.

iii-i:
I made the updates based on our discussion.

I also had a problem with python-docutils in my environment, which I used to solve with a private patch, but now I think it should be ok to publish it - "vmtest: use python3-docutils instead of python-docutils".

Travis support did not answer yet; unless I get an answer early next week, I think we should go with the libguestfs solution. Do you think whether it would make sense to completely replace the current image creation code with libguestfs one, or whether this should be s390x-only?

travis-ci/vmtest/run.sh
```diff
@@ -243,7 +243,7 @@ create_rootfs_img() {
 download_rootfs() {
 	local rootfsversion="$1"
 	local dir="$2"
-	download "${PROJECT_NAME}-vmtest-rootfs-$rootfsversion.tar.zst" |
+	download "$ARCH/${PROJECT_NAME}-vmtest-rootfs-$rootfsversion.tar.zst" |
```
wat-ze-hex, line 246: nit: `${ARCH}` for consistency?

wat-ze-hex:
@iii-i 

Could you please add environmental requirements for `mkrootfs_debian.sh` as a comment or amend it to commit message?
I've tried it in Ubuntu Bionic [have several distros in my VM and that's the closest to Debian ^^] and was able to build the image, though a few script modifications.

> I think we should go with the libguestfs solution. Do you think whether it would make sense to completely replace the current image creation code with libguestfs one, or whether this should be s390x-only?

I think that makes sense. The problem with `mkrootfs.sh` is that it is works-on-my-machine solution.

travis-ci/vmtest/mkrootfs_debian.sh
```diff
@@ -0,0 +1,26 @@
...
+	--cache-dir="$PWD/.debootstrap" \
+	--include=binutils,busybox,elfutils,iproute2,libcap2,libelf1,zlib1g \
+	--variant=minbase \
+	bullseye "$root"
+rm -rf \
+	"$root"/etc/rcS.d \
+	"$root"/usr/share/{doc,info,locale,man,zoneinfo} \
+	"$root"/var/cache/apt/archives/* \
+	"$root"/var/lib/apt/lists/*
+chroot "$root" dpkg --remove --force-remove-essential coreutils
```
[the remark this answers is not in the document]
iii-i: I'll double-check how this entire thing works on bionic, maybe will downgrade a few dependencies and document the rest.

It's kind of not too surprising that it fails this way, because `rm` is part of `coreutils`. However, the expectation here is that `coreutils` will be replaced with `busybox`.

travis-ci/vmtest/mkrootfs_tweak.sh
```diff
@@ -0,0 +1,61 @@
...
+cat > "$root/etc/inittab" << "EOF"
+::sysinit:/etc/init.d/rcS
+::ctrlaltdel:/sbin/reboot
+::shutdown:/sbin/swapoff -a
+::shutdown:/bin/umount -a -r
+::restart:/sbin/init
+EOF
+chmod 644 "$root/etc/inittab"
+
+mkdir -m 755 "$root/etc/init.d" "$root/etc/rcS.d"
```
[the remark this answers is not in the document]
iii-i: OK

travis-ci/vmtest/run.sh
iii-i: OK (I don't think `{}` does anything useful here, but I agree that it will look better).

travis-ci/vmtest/configs/INDEX
```diff
@@ -1,8 +1,9 @@
 INDEX	https://libbpf-vmtest.s3-us-west-1.amazonaws.com/x86_64/INDEX
-libbpf-vmtest-rootfs-2020.09.27.tar.zst	https://libbpf-vmtest.s3-us-west-1.amazonaws.com/x86_64/libbpf-vmtest-rootfs-2020.09.27.tar.zst
+x86_64/libbpf-vmtest-rootfs-2020.09.27.tar.zst	https://libbpf-vmtest.s3-us-west-1.amazonaws.com/x86_64/libbpf-vmtest-rootfs-2020.09.27.tar.zst
 vmlinux-4.9.0.zst	https://libbpf-vmtest.s3-us-west-1.amazonaws.com/x86_64/vmlinux-4.9.0.zst
 vmlinux-5.5.0-rc6.zst	https://libbpf-vmtest.s3-us-west-1.amazonaws.com/x86_64/vmlinux-5.5.0-rc6.zst
 vmlinux-5.5.0.zst	https://libbpf-vmtest.s3-us-west-1.amazonaws.com/x86_64/vmlinux-5.5.0.zst
 vmlinuz-5.5.0-rc6	https://libbpf-vmtest.s3-us-west-1.amazonaws.com/x86_64/vmlinuz-5.5.0-rc6
 vmlinuz-5.5.0	https://libbpf-vmtest.s3-us-west-1.amazonaws.com/x86_64/vmlinuz-5.5.0
 vmlinuz-4.9.0	https://libbpf-vmtest.s3-us-west-1.amazonaws.com/x86_64/vmlinuz-4.9.0
+s390x/libbpf-vmtest-rootfs-2021.03.24.tar.zst	https://github.com/iii-i/libbpf/blob/s390x-libbpf-vmtest-rootfs-2021.03.24/travis-ci/vmtest/libbpf-vmtest-rootfs-2021.03.24.tar.zst?raw=true
```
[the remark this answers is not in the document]
iii-i: OK

travis-ci/vmtest/mkrootfs_debian.sh
```diff
@@ -0,0 +1,26 @@
...
+set -e -u -x -o pipefail
+if [ "$(id -u)" != 0 ]; then
+	echo "$0 must run as root" >&2
+	exit 1
+fi
+root=$(mktemp -d -p "$PWD")
+trap 'rm -r "$root"' EXIT
+mkdir -p .debootstrap
+debootstrap \
+	--cache-dir="$PWD/.debootstrap" \
```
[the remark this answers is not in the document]
iii-i: OK

iii-i:
> > I think we should go with the libguestfs solution. Do you think whether it would make sense to completely replace the current image creation code with libguestfs one, or whether this should be s390x-only?
> 
> I think that makes sense. The problem with `mkrootfs.sh` is that it is works-on-my-machine solution.

I should've been a bit more precise here: I plan to rework only the part that copies selftests into the image in Travis, i.e. `run.sh`.

travis-ci/vmtest/configs/INDEX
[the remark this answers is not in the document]
wat-ze-hex: @iii-i 

Frankly, I copied https://github.com/iii-i/libbpf/blob/s390x-libbpf-vmtest-rootfs-2021.03.24/travis-ci/vmtest/libbpf-vmtest-rootfs-2021.03.24.tar.zst and tried the script on x86_64.
I would try it in VM.

iii-i:
I've added the libguestfs support and the mkrootfs bionic compatibility. Locally everything ran fine - let's see how it goes in Travis now.

iii-i:
The first problem was the old https://bugs.launchpad.net/fuel/+bug/1467579 bug that I circumvented with `sudo chmod +r /boot/vmlinuz*`. Now the problem is:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
```

I've added some debug output in the last few pushes, but this didn't shed any light. I will experiment with travis+libguestfs in a separate repo (for better turnaround times and to avoid spamming you :)) and report back.

anakryiko:
@iii-i is there anything else left to do on this change with Ubuntu Focal and other fixes landed?

iii-i:
Yeah, unfortunately it seems like we've opened yet another can of worms here. I made some further tests and it looks as if s390x Travis CI does not support KVM, which would make e.g. performance tests unacceptably slow. At the moment this doesn't look like a hard technical issue - nested virtualization is supported by both z/VM and KVM. We are currently trying to reach Travis regarding this.

anakryiko:
Oh, ok. Thanks for the update?

Do you think Github Actions would have the same problem? Maybe we should just switch to Github Actions? That's been on the low-pri TODO for a while anyways, so if you are going to investing in dealing with Travis CI so much, maybe it's better overall to just move to Github Actions?

iii-i:
I actually have investigating s390x + GitHub actions on my TODO list as well for another project (zlib-ng). But in the meantime we managed to convince Travis folks to look into this in greater detail, so let's wait for their findings.

iii-i:
At least some good news - we now have the s390x packages on apt.llvm.org. So I've rebased and removed the PPA commit. Ideally the build should pass now.

anakryiko:
@iii-i we are moving our CI jobs into Github Actions, see https://github.com/libbpf/libbpf/pull/323. Please check how that influences your work for s390x. Thanks.

iii-i:
As far as I can see, this now uses qemu-user, which is good for building, but not for running the tests - if we run `qemu-system-s390x` binary, it itself will run emulated, which we definitely don't want. And if we don't use https://github.com/uraimo/run-on-arch-action, then we'll have to cross-compile. Or we could come up with a hybrid setup, but I'd rather not go into this direction.

Travis folks have answered that they have no plans to fix the issues we are seeing here, so I tried to run the tests in emulation. There was a QEMU bug that I had to fix, but apart from that it worked. The problems are that 1) 20.04 backport that I requested will take forever and 2) it's quite slow.

So what do you think about creating a special configuration that will just SSH into the s390x virtual machine and invoke the scripts there? I thought about using https://github.com/actions/runner, which supports external builders, but it's not working on s390x at the moment.

anakryiko:
> So what do you think about creating a special configuration that will just SSH into the s390x virtual machine and invoke the scripts there? I thought about using https://github.com/actions/runner, which supports external builders, but it's not working on s390x at the moment.

If it's going to work reliably and won't need active maintenance, I'm ok with that.

iii-i:
@anakryiko After some QEMU TCG bugfixing, I've been able to install the x86_64 github actions runner to a s390x VM.

At the moment this is going to fail for two reasons:

* https://lore.kernel.org/bpf/20210906163635.302307-1-jean-philippe@linaro.org/
* The VM is registered as a builder for my libbpf fork, not for the upstream repo (obviously, I don't have permissions to do that). You can reconfigure it using the instructions that are part of this PR (https://github.com/iii-i/libbpf/blob/s390x/travis-ci/vmtest/s390x-self-hosted-builder/README.md). The builder has been moved to a better machine, I will send you the details by email.

travis-ci/vmtest/configs/BLACKLIST-latest.s390x
```diff
@@ -0,0 +1,45 @@
+# TEMPORARY
+atomics                                  # trampoline
+bpf_iter_setsockopt                      # kfunc
+bpf_tcp_ca                               # kfunc
+btf_dump                                 # ?
```
anakryiko, line 5: this is strange and should be fixed, let's not forget about it

travis-ci/vmtest/configs/BLACKLIST-latest.s390x
```diff
@@ -0,0 +1,45 @@
...
+core_autosize                            # fixed
```
anakryiko, line 6: fixed?

travis-ci/vmtest/configs/BLACKLIST-latest.s390x
```diff
@@ -0,0 +1,45 @@
...
+core_read_macros                         # BPF_PROBE_READ_KERNEL
+core_reloc                               # ?
```
anakryiko, line 8: same, it's troubling that CO-RE doesn't work as expected, let's look into that

anakryiko (commented):
I just see "This check failed" for LATEST + z15 build.

> You can reconfigure it using the instructions that are part of this PR 

Where those commands have to be run? You'll need to ELI5 me on this, if you want me to do something :)

anakryiko:
@thefallentree PTAL, Ilya is working on adding s390x as another architecture to BPF CI. It probably would useful if you review this and familiarize this in the face of ongoing BPF CI refactoring.

anakryiko:
> @anakryiko Sorry, I completely missed your comments - somehow I did not get any notifications in my mail. I will update the blacklist - I tried to triage all the issues, but it's still messy with all these "?"s and "fixed"s. IIRC most of the CO-RE stuff has to do with bitfield endianness. I'll follow up on that.

no worries

> 
> Regarding ELI5 - the first step would be to check that you can access the new server :) I sent you the new IP address privately on 8th of September.

Yeah, I do have an access. Next steps?

> I also noticed that you have https://github.com/libbpf/ci/ now - should I try to rebase on top of that, or would it be better to continue here?

We do, and we are in a process of consolidating all the variations of CI scripts (@thefallentree is on it), but for now just stick to libbpf CI scripts in this repo, no need to rebase as we don't use libbpf/ci files here yet.

iii-i:
In order to attach the builder to https://github.com/libbpf/libbpf, you need to go to https://github.com/libbpf/libbpf/settings/actions/runners/new (I should probably add this link to the doc) and copy the token from the "Configure" section.

Then on the server do the following (come to think of it, this is close to the doc, but is never spelled out in full there):
```
$ sudo systemctl stop actions-runner
$ sudo docker run \
      --rm \
      --interactive \
      --tty \
      --volume=actions-runner:/home/actions-runner \
      iiilinuxibmcom/actions-runner-libbpf \
      ./config.sh --url https://github.com/libbpf/libbpf --token <token>
$ sudo systemctl start actions-runner
```

After a minute you should see the builder in https://github.com/libbpf/libbpf/settings/actions/runners.

iii-i:
It seems to fully work for me now: https://github.com/iii-i/libbpf/pull/1. Changes since the last time:

- `--ephemeral` GitHub Actions Runner support.
- Use `next` pahole branch.
- Update `5.5.0` blacklist.

Regarding what needs to be done to link the builder with the upstream repo (this is now documented, but I'm pasting it here for convenience):

- Update `/etc/actions-runner-libbpf` (`access_token` should have `repo` scope, I created mine at https://github.com/settings/tokens)
- Delete persistent data and restart the runner:

```
$ sudo systemctl stop actions-runner-libbpf
$ sudo docker rm -f actions-runner-libbpf
$ sudo docker volume rm actions-runner-libbpf
$ sudo systemctl start actions-runner-libbpf
```

anakryiko (commented):
oops, seems like I forgot to submit my comments

anakryiko:
I did enable the test runner and it seems to be working fine! Please address few small nits and undo the BLACKLIST change, and this should be good to go.

.github/workflows/test.yml
```diff
@@ -30,6 +35,7 @@ jobs:
         name: vmtest
         with:
           kernel: ${{ matrix.kernel }}
+          pahole: next
```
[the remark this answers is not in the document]
iii-i: My fixes for btf_dump were only in next. I thought that since we already use latest-greatest llvm and kernel, doing the same for pahole won't hurt. But in the meantime my patches propagated to master, so let's drop this change.

travis-ci/vmtest/configs/BLACKLIST-5.5.0
```diff
@@ -3,20 +3,25 @@ align			# verifier output format changed
 atomics			# new atomic operations (v5.12+)
 atomic_bounds		# new atomic operations (v5.12+)
 bind_perm		# changed semantics of return values (v5.12+)
+bloom_filter_map	# v5.16+
```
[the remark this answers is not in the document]
iii-i: Ah, I see now why it worked upstream. I'll check why whitelisting is not working anymore with my changes.

travis-ci/vmtest/run_selftests.sh
```diff
@@ -1,9 +1,17 @@
 #!/bin/bash
 
-set -euo pipefail
+set -euxo pipefail
```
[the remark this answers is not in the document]
iii-i: OK.

travis-ci/vmtest/run_selftests.sh
```diff
@@ -30,16 +38,11 @@ test_verifier() {
...
-
-whitelist_path="$configs_path/whitelist/WHITELIST-${KERNEL}"
-if [[ -s "${whitelist_path}" ]]; then
-	WHITELIST=$(cat "${whitelist_path}" | cut -d'#' -f1 | tr -s '[:space:]' ',')
-fi
+configs_path=libbpf/travis-ci/vmtest/configs
+BLACKLIST=$(read_lists "$configs_path/BLACKLIST-${KERNEL}" \
+		       "$configs_path/BLACKLIST-${KERNEL}.${ARCH}")
+WHITELIST=$(read_lists "$configs_path/WHITELIST-${KERNEL}" \
+		       "$configs_path/WHITELIST-${KERNEL}.${ARCH}")
```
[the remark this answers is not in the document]
iii-i: OK.

travis-ci/vmtest/run.sh
```diff
@@ -368,23 +393,22 @@ if [[ -v BUILDDIR || $ONESHOT -eq 0 ]]; then
...
-	echo "Copying vmlinux..." >&2
-	sudo rsync -cp --chmod 0644 "$source_vmlinux" "$vmlinux"
 else
-	# We could use "sudo zstd -o", but let's not run zstd as root with
-	# input from the internet.
-	download "vmlinux-${KERNELRELEASE}.zst" |
-		zstd -d | sudo tee "$vmlinux" > /dev/null
-	sudo chmod 644 "$vmlinux"
+	source_vmlinux="${ARCH_DIR}/vmlinux-${KERNELRELEASE}"
+	download "vmlinux-${KERNELRELEASE}.zst" | zstd -d >"$source_vmlinux"
```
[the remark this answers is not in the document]
iii-i: Isn't it better to do it the other way around and use `${ARCH_DIR}` for the rootfs image too? This way all temporary files will be placed according to `-d` option.

iii-i:
- Dropped the pahole change.
- Fixed whitelists.
- Added $ARCH to vmlinux names.
- Dropped -x change.
- Small style improvements.

anakryiko:
@iii-i, looks great, I'm merging it, thanks! Note that there is currently one compiler warning on s390x, probably worth looking into it and fixing to keep the compilation completely clean:

```
  progs/test_probe_user.c:40:43: warning: passing 'const volatile unsigned long *' to parameter of type 'const void *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
          bpf_probe_read_kernel(&ptr, sizeof(ptr), &PT_REGS_PARM2(real_regs));
                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~
  1 warning generated.
```

@thefallentree, please keep these changes in mind when consolidating BPF CI logic, we shouldn't lose all these changes during the migration process.

thefallentree:
> @thefallentree, please keep these changes in mind when consolidating BPF CI logic, we shouldn't lose all these changes during the migration process.

ack

PR #212, opened 2020-09-29
update bpf_stats_type declaration as c++11 Unscoped enumerations

b-ripper:
when bcc reference bpf.h, compail will has erroe:
file ISO C++ forbids forward references to 'enum' types

Signed-off-by: eason<eason.sheng.chen@gmail.com>


anakryiko:
Please submit the patch to bpf@vger.kernel.org. This repo is mirrored from in-kernel libbpf sources, so patches like this can't be directly applied here.

PR #232, opened 2021-01-26
install: don't preserve file owner

teknoraver:
'cp -p' preserve file ownership, this may leave files owned by the
current in user in /lib .

Signed-off-by: Matteo Croce <mcroce@microsoft.com>

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -130,7 +130,7 @@ define do_s_install
 	$(Q)if [ ! -d '$(DESTDIR)$2' ]; then		\
 		$(INSTALL) -d -m 755 '$(DESTDIR)$2';	\
 	fi;
-	$(Q)cp -fpR $1 '$(DESTDIR)$2'
+	$(Q)cp -fR $1 '$(DESTDIR)$2'
 endef
 
 install: all install_headers install_pkgconfig


rdna:
I think it should be fine.

`-p` preserves three things: mode, ownership, timestamps:
* ownership should be fine: files will be installed with user:group of whoever runs the install;
* timestamps should be fine to ignore since other places in this `Makefile` use `install` and don't preserve them either;

And mode is the most important part since e.g. for `.so` execution bit must be preserved. From what I see `cp(1)` uses src mode as `mode` arguments for `open(2)` when opening dst file [0]. And POSIX says same thing (see [1]):
> 3. If source_file is of type regular file, the following steps shall be taken:
>   b. If dest_file does not exist, a file descriptor shall be obtained by performing actions equivalent to the open() function defined in the System Interfaces volume of POSIX.1-2017 called using dest_file as the path argument, and the bitwise-inclusive OR of O_WRONLY and O_CREAT as the oflag argument. The file permission bits of source_file shall be the mode argument.

So it seems mode should still be transferred from src to dst for regular files. Form symlinks they should not matter according to `symlink(2)`.





[0] https://github.com/coreutils/coreutils/blob/32ff1d1313ff566685b9092ac929251a4c22e7c4/src/copy.c#L2912-L2922
[1] https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cp.html

PR #245, opened 2021-03-05
Sync floating-point support from the kernel

iii-i:
This is not a full bpf-next sync, just up to 22541a9eeb0d. The changes that follow create merge conflicts in Makefile/bpf_helpers_doc.py, so I've decided against including them here.


anakryiko:
Thanks! BTW, how was your experience with the sync patch?

iii-i:
It's quite good I must say. The only hiccup was that I was trying to sync a kernel commit which was not checked out (like this: `scripts/sync-kernel.sh $PWD ~/myrepos/linux 22541a9eeb0d968c133aaebd95fa59da3208e705`), and the checked out one did not have `scripts/bpf_helpers_doc.py`. But moving scripts around doesn't happen frequently I think.

anakryiko:
Oh, right, yeah, we'll need to adjust scripts to accommodate new files like that. I'll get to it with the next sync.

PR #250, opened 2021-03-24
sync: latest libbpf changes from kernel

iii-i:
Hi,

I made another sync in order to pick up the s390 atomics changes.
There were a few problems with changes in the kernel Build/Makefile that had to be manually transplanted to the libbpf Makefile, as well as the bpf_helper_doc.py -> bpf_doc.py renaming, but I think I got everything right in the end.


anakryiko:
Seems like all the tests are failing due to invalid GIT hash:

```
fatal: Not a valid object name d067cfe05eb9c863ecd6807c8e771624974e3efe^{commit}
fatal: Could not parse object 'd067cfe05eb9c863ecd6807c8e771624974e3efe'.
```

Can you please re-sync and update the PR? Thanks.

iii-i:
Seems to work now. I also realized I missed the `libbpf_util.h` removal - this is also fixed.

anakryiko:
Ok, so this doesn't look quite right... Both BPF-CHECKPOINT-COMMIT and CHECKPOINT-COMMIT point to the same SHA, which is highly suspicious. They are supposed to point to bpf and bpf-next trees, respectively. Currently bpf and bpf-next have different top commits, so you should end up with two different SHAs, unless I missed something.

How are you doing this sync, btf? I run it as such:
```
scripts/sync-kernel.sh ~/libbpf ~/linux bpf-master
```
I do that from linux repo, checked out on bpf-next's master. bpf-master branch points to latest bpf/master.

Can you please share the full log of sync-kernel.sh output?

iii-i:
Ah, I see my mistake - I used `bpf-next/master` instead of `bpf/master`. Let me try one more time..

anakryiko:
No worries, no one else yet used that script beside me :) Though I think I got it to the point where most of the time it just works. Please check that there are either no differences between contents of linux and libbpf sides or differences are only due to bpf-master diffs, that are not yet merged into bpf-next tree. For the first time please just put entire log into Git Gist, I'll double-check as well.

iii-i:
Log: https://pastebin.com/jHE5QDj1
It must have failed early due differing file lists, but at the end it seems to be okay: `libbpf$ colordiff -u -r src/ ~/myrepos/linux/tools/lib/bpf` shows differences only in `Makefile` and `.gitignore`.
And this time there is 1 more commit - presumably from the bpf branch.
Btw, I think I messed it up in https://github.com/libbpf/libbpf/pull/245 time too, but now everything looks okay?

anakryiko:
@iii-i I don't know what you did differently, but I got absolutely no differences and no patches that needs manual resolution. See my log at https://gist.github.com/anakryiko/a6c4121fda2e45a3e3fc1b6094482046, I only did bpf_doc.py change locally for that. But you are right, I noticed that https://github.com/libbpf/libbpf/pull/245 ended up with the same SHA, so I want to rerun sync from bpf tree sha before the last sync, just in case we are missing anything (and to test that script really handles duplicates well).

Also, fexit_sleep test needs to be blacklisted now on 5.5 kernel, it relies on the very latest fix from bpf tree. Let me run similar sync and post a separate PR, so that you can compare, if you want to.

iii-i:
What part of `sync-kernel.sh` is supposed to filter out `Makefile` and `Build` changes? I see there is `LIBBPF_PATHS`, but it seems to be used only to skip the entire commits and to resolve conflicts when working in the `linux` tree. In my case `${TMP_DIR}/patches` contained changes to `Makefile` and `Build`, and that's what required resolution.

anakryiko:
See LIBBPF_TREE_FILTER, which is used to rewrite Linux repo to match Github layout. It skips Makefile and Build. So I missed one more Makefile change (adding strset.o and linker.o). But it's more of a problem of testing, in our CI runs that runs selftests we actually build libbpf from linux sources, not from Github sources. It should ideally be built from Github and linked into selftests, I just can't bring myself to changing that.

We do have just build tests, but those don't catch missing .o, because you need to use symbols from those .o files to see linker failure :( Sadness. But doesn't happen often, so I haven't prioritized it.

PR #235, opened 2021-02-02
travis-ci: update to Ubuntu Focal

anakryiko:
Update to Ubuntu Focal 20.04.

Signed-off-by: Andrii Nakryiko <andrii@kernel.org>

PR #252, opened 2021-03-30
travis-ci: update to Ubuntu Focal [v2.0]

iii-i:
@anakryiko Seems like adding libguestfs for https://github.com/libbpf/libbpf/pull/204 requires moving to focal, because bionic libguestfs uses bionic qemu-system-s390x, which is too old for the modern s390x kernels (I remember I bumped into it during my first attempt to add s390x to the CI). I've experimented with the focal libguestfs in a separate repo, and it worked there. So I've taken your commit from https://github.com/libbpf/libbpf/pull/235 for this PR (without the pahole part) and added two fixes on top.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 sudo: required
 language: bash
-dist: bionic
+dist: focal
 services:
     - docker
 
@@ -87,21 +87,21 @@ jobs:
           script:         $CI_ROOT/managers/debian.sh RUN_GCC10_ASAN || travis_terminate 1
           after_script:   $CI_ROOT/managers/debian.sh CLEANUP
 
-        - name: Ubuntu Bionic Build
+        - name: Ubuntu Focal Build
           language: bash
           script: sudo $CI_ROOT/managers/ubuntu.sh || travis_terminate 1
 
-        - name: Ubuntu Bionic Build (arm)
+        - name: Ubuntu Focal Build (arm)
           arch: arm64
           language: bash
           script: sudo $CI_ROOT/managers/ubuntu.sh || travis_terminate 1
 
-        - name: Ubuntu Bionic Build (s390x)
+        - name: Ubuntu Focal Build (s390x)
           arch: s390x
           language: bash
           script: sudo $CI_ROOT/managers/ubuntu.sh || travis_terminate 1
 
-        - name: Ubuntu Bionic Build (ppc64le)
+        - name: Ubuntu Focal Build (ppc64le)
           arch: ppc64le
           language: bash
           script: sudo $CI_ROOT/managers/ubuntu.sh || travis_terminate 1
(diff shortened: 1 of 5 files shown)


iii-i:
There is still one problem remaining:

```
#112 stacktrace_build_id:FAIL
#113 stacktrace_build_id_nmi:OK
```

This is my understanding of how `stacktrace_build_id` is supposed to work. We attach a BPF program to the `urandom_read` tracepoint and then trigger urandom accesses from the userspace using `dd` and `./urandom_read`. Then we expect to see `./urandom_read`'s build ID in the stacktrace gathered by the BPF program.

However, it's not `./urandom_read` itself that issues the syscall that triggers the tracepoint, but rather the libc. So we rely on the ability of the kernel to unwind libc's stack frame(s). Is this a realistic expectation? As far as I understand, the x86_64 kernel can only follow the frame pointers, and if they are compiled out, then we are out of luck. And while we can control how we are building the test code, we can't control how the system libc is built.

`stacktrace_build_id_nmi` on the other hand looks okay, because it attaches the BPF program to the `PERF_COUNT_HW_CPU_CYCLES` perf event, and at least one of those should hit the `./urandom_read` binary itself.

test_stacktrace_build_id is in `BLACKLIST-latest.s390x` exactly because of this reason. On s390x the situation is more dramatic than on x86_64 though, because I don't think we ever built the userspace with "backchains". But seems like we now have more or less the same situation on x86_64: the only reasonable way to get stacktraces is to save the raw stack and ask something in userspace to parse it using DWARF.

Given all that, I'm not sure how to reasonably fix the test. We could, of course, write inline syscall wrappers for all the supported architectures, but that doesn't sounds nice at all.

iii-i:
Turns out this has to do with build id parsing. Here is the `urandom_read` program header when we compile it on Ubuntu 20.04:

```
Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .interp           PROGBITS         0000000000000318  00000318
       000000000000001c  0000000000000000   A       0     0     1
  [ 2] .note.gnu.propert NOTE             0000000000000338  00000338
       0000000000000020  0000000000000000   A       0     0     8
  [ 3] .note.gnu.build-i NOTE             0000000000000358  00000358
       0000000000000024  0000000000000000   A       0     0     4
```

v5.5 kernel does this:

```
        for (i = 0; i < ehdr->e_phnum; ++i)
                if (phdr[i].p_type == PT_NOTE)
                        return stack_map_parse_build_id(page_addr, build_id,
                                        page_addr + phdr[i].p_offset,
                                        phdr[i].p_filesz);
```

So it finds `.note.gnu.propert` and bails. LATEST is a bit more cautious:

```
        for (i = 0; i < ehdr->e_phnum; ++i) {
                if (phdr[i].p_type == PT_NOTE &&
                    !parse_build_id(page_addr, build_id, size,
                                    page_addr + phdr[i].p_offset,
                                    phdr[i].p_filesz))
                        return 0;
        }
``` 

thanks to

```
commit b33164f2bd1cedb094c32cb466287116164457ae
Author: Jiri Olsa <jolsa@kernel.org>
Date:   Wed Aug 12 14:31:02 2020 +0200

    bpf: Iterate through all PT_NOTE sections when looking for build id
```

from v5.9. So I guess we should just blacklist the test for v5.5.

anakryiko:
Awesome, thanks!

PR #263, opened 2021-04-28
vmtest: use clang-14 as latest nightly

anakryiko:
Seems like nightly got switched tonight to clang-14.


anakryiko:
false assumption, abandoning, fixed by https://github.com/libbpf/libbpf/pull/264

PR #322, opened 2021-06-11
Pin pahole version to known-working

tsipa:
Seems like ae2581647e84948810ba209f3891359dd4540110  breaks build

Signed-off-by: Sergei Iudin <siudin@fb.com>


anakryiko:
I think for libbpf CI it's fine to wait for pahole master to get fixed (it doesn't matter if build goes green if I don't sync anything). kernel-patches CI is a different matter, though, so we should pin pahole there for sure.

Issue #316, opened 2021-06-02
travis-ci: migrate libbpf CI from Travis CI to Github Actions


anakryiko:
We need to migrate libbpf CI from the current setup with Travis CI to something like Github Actions to avoid paying for Travis CI ($69 per month) while getting only single build at a time. From the looks of it, Github Actions is both free and would provide better and faster CI capabilities.

https://github.com/libbpf/libbpf/blob/master/.travis.yml is the starting point.

tohojo:
We have the same issue for xdp-tools (cf https://github.com/xdp-project/xdp-tools/issues/114) - linking the issues to subscribe to any updates here :)

tsipa:
https://github.com/tsipa/vmtest/commit/d170ddc185e539f40855a7838916d4775708129b i did necessary changes to make syntax compatible with gh actions, but gh will not let us run qemu  inside their containers.

The explanation is available here https://github.com/actions/virtual-environments/issues/183
TL;DR "azure something-something, github something-something, we won't support it in a human life".


I don't understand the subject area to justify if we can run these tests inside the docker container wo VM, but if no - we have to consider something else, not GH actions.

tohojo:
I did some experiments with getting Qemu to run in software-emul mode here: https://github.com/xdp-project/xdp-tools/pull/113

It runs on Travis, but haven't tried converting it to GH actions; care to try that out? It's just adding a couple of qemu args (`-machine accel=tcg -cpu qemu64`) so should be simple enough to adapt, you'll just need to get rid of the kvm stuff of course...

tsipa:
Thanks for this work, will try be eod as i already have dev env set up.

But i strongly suspect without hardware virtualization it's going to be a torture and we likely to hit many weird issues caused by inadequate timeouts here and there - SW emulation is really slow.

tohojo:
"Sergey V. Yudin" ***@***.***> writes:

> Thanks for this work, will try be eod as i already have dev env set
> up.

Awesome!

> But i strongly suspect without hardware virtualization it's going to
> be a torture and we likely to hit many weird issues caused by
> inadequate timeouts here and there - SW emulation is really slow.

Yeah, but maybe that'll be an excuse to rely less on timeouts, and thus
get more reliable tests? One can hope, at least ;)

tsipa:
Wow, it worked and it wasn't that bad:
travis 32min
https://travis-ci.com/github/kernel-patches/bpf/builds/228452123
GH actions 36min
[...]

PR #323, opened 2021-06-15
Duplicate PR CI with github actions

tsipa:
https://github.com/libbpf/libbpf/issues/316

This change supposed to have backward compatibility with travis and with this change i'm expecting that we will run two CIs at the same time.

I would like to get some assistance on coverity - it would be nice if someone can go though https://docs.github.com/en/actions/reference/encrypted-secrets this and set up token

However i can foresee security implications of this - i don't understand what would prevent users from making a PR which would change workflow to print out security token.

### DIFF
--- a/.github/actions/vmtest/action.yml
+++ b/.github/actions/vmtest/action.yml
@@ -1,0 +1,15 @@
+name: 'vmtest'
+description: 'Build + run vmtest'
+inputs:
+  kernel:
+    description: 'kernel or LATEST'
+    required: true
+    default: 'LATEST'
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        source /tmp/ci_setup
+        export KERNEL=${{ inputs.kernel }}
+        $CI_ROOT/vmtest/run_vmtest.sh
+      shell: bash
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,85 @@
+name: libbpf-ci
+
+on: 
+  pull_request:
+  push:
+  schedule:
+    - cron:  '0 18 * * *'
+
+
+jobs:
+  vmtest:
+    runs-on: ubuntu-latest
+    name: Kernel ${{ matrix.kernel }} + selftests
+    strategy:
+      matrix:
+        include:
+          - kernel: 'LATEST'
+          - kernel: '5.5.0'
+          - kernel: '4.9.0'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/setup
+      - uses: ./.github/actions/vmtest
+        with:
+          kernel: ${{ matrix.kernel }}
+
+  debian:
+    runs-on: ubuntu-latest
+    name: Debian Build (${{ matrix.name }})
+    strategy:
+      matrix:
+        include:
+          - name: default
+            target: RUN
+          - name: ASan+UBSan
+            target: RUN_ASAN
+          - name: clang
+            target: RUN_CLANG
+          - name: clang ASan+UBSan
+            target: RUN_CLANG_ASAN
+          - name: gcc-10
+            target: RUN_GCC10
+          - name: gcc-10 ASan+UBSan
+            target: RUN_GCC10_ASAN
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/setup
+      - uses: ./.github/actions/debian
+        with:
+          target: ${{ matrix.target }}
+
+  ubuntu:
+    runs-on: ubuntu-latest
+    name: Ubuntu Focal Build (${{ matrix.arch }})
+    strategy:
+      matrix:
+        include:
+          - arch: aarch64
+          - arch: ppc64le
+          - arch: s390x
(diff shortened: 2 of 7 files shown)


.github/actions/vmtest/action.yml
```diff
@@ -0,0 +1,17 @@
...
+    description: 'kernel or LATEST'
+    required: true
+    default: 'LATEST'
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        ls -la /tmp/
+        cat /tmp/ci_setup
+        . /tmp/ci_setup
```
anakryiko, line 14: nit: '.' is easy to miss, `source /tmp/ci_setup` would be a bit more obvious

.github/workflows/test.yml
```diff
@@ -0,0 +1,86 @@
+name: bpfci
```
anakryiko, line 1: libbpf-ci?

.github/workflows/test.yml
```diff
@@ -0,0 +1,86 @@
...
+            cp /tmp/ci_setup $GITHUB_WORKSPACE
+          dockerRunArgs: |
+              --volume "${GITHUB_WORKSPACE}:${GITHUB_WORKSPACE}"
+          shell: /bin/bash
+          install: |
+              export DEBIAN_FRONTEND=noninteractive
+              export TZ="America/Los_Angeles"
+              apt-get update -y
+              apt-get install -y tzdata
+              apt-get install -y build-essential sudo
```
anakryiko, line 85: nit: can combine into a single command?

travis-ci/vmtest/build_pahole.sh
```diff
@@ -15,7 +15,8 @@ cd ${REPO_PATH}
 git init
 git remote add origin ${PAHOLE_ORIGIN}
 git fetch origin
-git checkout master
+# git checkout master
+git checkout 743f2536d8b876a4d9addc6d6d9eb1aa086a47c8
```
anakryiko, line 19: pahole got fixed, so we don't need this anymore

travis-ci/vmtest/run.sh
```diff
@@ -436,11 +436,17 @@ sudo umount "$mnt"
...
+if kvm-ok ; then
+  qemu-system-x86_64 -nodefaults -display none -serial mon:stdio \
+    -cpu kvm64 -enable-kvm -smp "$(nproc)" -m 4G \
+    -drive file="$IMG",format=raw,index=1,media=disk,if=virtio,cache=none \
+    -kernel "$vmlinuz" -append "root=/dev/vda rw console=ttyS0,115200$APPEND"
+else
+  qemu-system-x86_64 -nodefaults -display none -serial mon:stdio \
+	  -cpu qemu64 -machine accel=tcg -smp "$(nproc)" -m 4G \
+  	-drive file="$IMG",format=raw,index=1,media=disk,if=virtio,cache=none \
+	  -kernel "$vmlinuz" -append "root=/dev/vda rw console=ttyS0,115200$APPEND"
```
anakryiko, line 448: Oh, that's a lot of copy/paste of a very long list of arguments... Why not extract what differs into a variable. Something along these lines:

```
if kvm-ok then
QEMU_CPU_ARGS=-cpu kvm64 -enable-kvm
else
QEMU_CPU_ARGS=-cpu qemu64 -machine accel=tcg
fi
qemu-system-x86_64 .... $(QEMU_CPU_ARGS) ....
```
?

anakryiko (commented):
Super excited to finally have an alternative to Travis CI!

I don't see any Github Actions runs on this PR. Is there any way to check that before landing PR?

As for the secrets, I can set up what's necessary. Do you know what exactly needs to be done or were you asking for someone to figure it out? But seems like we just hard-code some token in .travis.yaml right now, so if we do the same for Github Actions it won't be a regression :) So we can just start with that, I suppose?

tsipa:
> I don't see any Github Actions runs on this PR. Is there any way to check that before landing PR?
It worked in my repo:
https://github.com/tsipa/libbpf/actions

but i dunno how to trigger CI against this. Maybe that is the answer to my security concern - the changes to CI not reflected on PR.

> As for the secrets, I can set up what's necessary. Do you know what exactly needs to be done or were you asking for someone to figure it out?
I need to have secure variable to be created in this repo and i think i will be able to figure out the rest.

>But seems like we just hard-code some token in .travis.yaml right now, so if we do the same for Github Actions it won't be a regression :) So we can just start with that, I suppose?
I don't know the right answer to this question. If i understand correctly if i do changes to CI code e.g. add echo $SECURE_VAR and send PR with this i would be able to expose secure variable, but mb i don't understand something.

anakryiko:
Great, thanks a lot for working on this.

evverx:
FWIW to judge from https://github.com/evverx/libbpf-fork/pull/2 (where the code should have failed to compile due a missing semicolon) the "vmtest" workflow doesn't test the code on GitHub so it would probably make sense to not run it on PRs and commits and just keep the "cron" job to make sure it's safe to sync the libbpf repository with the kernel tree I assume.

PR #324, opened 2021-06-15
Add coverity workflow

tsipa:
I've added COVERITY_SCAN_TOKEN and this change should work.
Seems like we have limit for scans and now it fails with:
"Coverity Scan analysis NOT authorized until Wed, 16 Jun 2021 01:04:24 +0000 UTC."
https://github.com/tsipa/libbpf/runs/2833993416?check_suite_focus=true

but previously i've got green builds: https://github.com/tsipa/libbpf/actions/runs/940726229

also removed 
"echo travis_fold:start:shutdown"
which doesn't have an end and it's not clear what is the purpose of this. If you know that it's intentionally put here pls let me know - i'll rework this part.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,0 +1,30 @@
+name: libbpf-ci-coverity
+
+on:
+  schedule:
+    - cron:  '0 18 * * *'
+
+
+jobs:
+  coverity:
+    runs-on: ubuntu-latest
+    name: Coverity
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/setup
+      - name: Run coverity
+        run: |
+          echo ::group::Setup CI env
+          source /tmp/ci_setup
+          export COVERITY_SCAN_NOTIFICATION_EMAIL="${AUTHOR_EMAIL}"
+          export COVERITY_SCAN_BRANCH_PATTERN=${GITHUB_REF##refs/*/}
+          export TRAVIS_BRANCH=${COVERITY_SCAN_BRANCH_PATTERN}
+          echo ::endgroup
+          scripts/coverity.sh
+        env:
+          COVERITY_SCAN_TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
+          COVERITY_SCAN_PROJECT_NAME: libbpf
+          COVERITY_SCAN_BUILD_COMMAND_PREPEND: 'cd src/'
+          COVERITY_SCAN_BUILD_COMMAND: 'make'
+      - name: SCM log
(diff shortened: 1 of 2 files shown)


travis-ci/vmtest/run.sh
```diff
@@ -425,7 +425,6 @@ sudo chmod 755 "$mnt/etc/rcS.d/S50-run-tests"
 
 poweroff_script="#!/bin/sh
 
-echo travis_fold:start:shutdown
```
anakryiko, line 428: it's supposed to match https://github.com/libbpf/libbpf/pull/324/files#diff-0b57a502f172fb2b6b4d53d5c3f0f9154af42788f3444c1224ab1d3b51d9ce62L457, but we don't have travis_fold function inside the VM.. I see how that's painful now, so not sure what to do about it. We can drop it or just add (uncoditionally) both travis and github actions-style section start output

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,30 @@
...
+          export TRAVIS_BRANCH=${COVERITY_SCAN_BRANCH_PATTERN}
+          echo ::endgroup
+          scripts/coverity.sh
+        env:
+          COVERITY_SCAN_TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
+          COVERITY_SCAN_PROJECT_NAME: libbpf
+          COVERITY_SCAN_BUILD_COMMAND_PREPEND: 'cd src/'
+          COVERITY_SCAN_BUILD_COMMAND: 'make'
+      - name: SCM log
+        run: cat /home/runner/work/libbpf/libbpf/src/cov-int/scm_log.txt
```
anakryiko, line 30: seems like there are a lot of unrelated warnings, do you think we can avoid them somehow? Coverity has ability to specify file patterns to ignore or categorize errors, it could be a way to make those logs clean. Either way, just something I noticed (I didn't see those scm_logs.txt before).

tsipa: I've seen them and decided not to investigate too deep as it's orthogonal to GH actions migration - it seems like we have same behavior for travis, nobody just bothered looking at this log.

travis-ci/vmtest/run.sh
```diff
@@ -423,9 +423,10 @@ fi
 echo "${setup_script}" | sudo tee "$mnt/etc/rcS.d/S50-run-tests" > /dev/null
 sudo chmod 755 "$mnt/etc/rcS.d/S50-run-tests"
 
+fold_sutdown="$(travis_fold start shutdown)"
```
anakryiko, line 426: oh, clever! typo though: sutdown -> shutdown

PR #341, opened 2021-07-19
Run apt-get update as a first step for GH actions

tsipa:
otherwise container may contain stall repo metadata cached


anakryiko:
Awesome, it fixed all the known-good tests. Thanks!

PR #345, opened 2021-07-30
Improve CI UX a little by setting names a hiding debug

tsipa:


anakryiko:
So much better, thanks!

PR #348, opened 2021-08-04
Fix text grouping issue on github actions

thefallentree:
github action grouping is broken because we were outputting "::endgroup" where
it needs "::endgroup::". This patch also added some additional grouping around
container setup phase, making output easier to read.


thefallentree:
compare this run with previous run:
image

image

anakryiko (approved):
Very nice, thank you!

PR #349, opened 2021-08-04
Add a test step to produce a minimal binary using libbpf.

thefallentree:
This patch adds a test step to link a minimal program to libbpf library produced,
making sure that the library works.

### DIFF
--- a/travis-ci/managers/debian.sh
+++ b/travis-ci/managers/debian.sh
@@ -48,6 +48,7 @@ for phase in "${PHASES[@]}"; do
             echo -e "::endgroup::"
             ;;
         RUN|RUN_CLANG|RUN_GCC10|RUN_ASAN|RUN_CLANG_ASAN|RUN_GCC10_ASAN)
+            CC="cc"
             if [[ "$phase" = *"CLANG"* ]]; then
                 ENV_VARS="-e CC=clang -e CXX=clang++"
                 CC="clang"
@@ -62,7 +63,7 @@ for phase in "${PHASES[@]}"; do
                 CFLAGS="${CFLAGS} -fsanitize=address,undefined"
             fi
             docker_exec mkdir build install
-            docker_exec ${CC:-cc} --version
+            docker_exec ${CC} --version
             info "build"
             docker_exec make -j$((4*$(nproc))) CFLAGS="${CFLAGS}" -C ./src -B OBJDIR=../build
             info "ldd build/libbpf.so:"
@@ -73,7 +74,8 @@ for phase in "${PHASES[@]}"; do
             fi
             info "install"
             docker_exec make -j$((4*$(nproc))) -C src OBJDIR=../build DESTDIR=../install install
-            docker_exec rm -rf build install
+            info "link binary"
+            docker_exec bash -c "CFLAGS=\"${CFLAGS}\" ./travis-ci/managers/test_compile.sh"
             ;;
         CLEANUP)
             info "Cleanup phase"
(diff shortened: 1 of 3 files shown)


travis-ci/managers/debian.sh
```diff
@@ -73,7 +74,13 @@ for phase in "${PHASES[@]}"; do
             fi
             info "install"
             docker_exec make -j$((4*$(nproc))) -C src OBJDIR=../build DESTDIR=../install install
-            docker_exec rm -rf build install
+            info "link binary"
+            docker_exec bash -c "echo -e '#include <bpf/libbpf.h> \n int main() { return bpf_object__open(0) == 0; }' | ${CC} ${CFLAGS} -o main -xc -L./install/usr/lib64 -L./install/usr/lib -I./install/usr/include - -lbpf"
```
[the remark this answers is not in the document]
thefallentree: done

thefallentree:
PTAL: 

note this already uncovered a issue in the Makefile: the static library is missing recently added `relo_core.o` file

rafaeldtinoco:
https://github.com/libbpf/libbpf/pull/353

travis-ci/managers/ubuntu.sh
```diff
@@ -19,5 +19,5 @@ if ! ldd build/libbpf.so | grep -q libelf; then
     echo "FAIL: No reference to libelf.so in libbpf.so!"
     exit 1
 fi
-make -j$((4*$(nproc))) -C src OBJDIR=../build DESTDIR=../install install
-rm -rf build install
+make -j$((4*$(nproc))) -C src V=1 OBJDIR=../build DESTDIR=../install install
```
anakryiko, line 22: do you still need V=1 or it's just debug leftover?

src/Makefile
```diff
@@ -36,7 +36,7 @@ SHARED_OBJDIR := $(OBJDIR)/sharedobjs
 STATIC_OBJDIR := $(OBJDIR)/staticobjs
 OBJS := bpf.o btf.o libbpf.o libbpf_errno.o netlink.o \
 	nlattr.o str_error.o libbpf_probes.o bpf_prog_linfo.o xsk.o \
-	btf_dump.o hashmap.o ringbuf.o strset.o linker.o gen_loader.o
+	btf_dump.o hashmap.o ringbuf.o strset.o linker.o gen_loader.o relo_core.o
```
anakryiko, line 39: I've applied a separate PR with a fix, please rebase

anakryiko (approved):
Looks good, thanks! Please rebase and drop V=1 and let's apply this.

src/Makefile
thefallentree: done!

travis-ci/managers/ubuntu.sh
thefallentree: removed.

anakryiko (changes requested):
Actually, see the log:
```
+ ldd main_shared
	linux-vdso.so.1 (0x00007ffc3f35a000)
	libbpf.so.0 => not found
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8288815000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f82889e7000)
```

I don't think shared library test is actually testing anything... So I don't know, let's either remove it, or we'd need to set everything correctly and run it.

thefallentree:
> Actually, see the log:
> 
> ```
> + ldd main_shared
> 	linux-vdso.so.1 (0x00007ffc3f35a000)
> 	libbpf.so.0 => not found
> 	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8288815000)
> 	/lib64/ld-linux-x86-64.so.2 (0x00007f82889e7000)
> ```
> 
> I don't think shared library test is actually testing anything... So I don't know, let's either remove it, or we'd need to set everything correctly and run it.

The binary did link correctly, but I think it doesn't show up in ldd because we are not setting runtime path for libbpf.so correctly for some reason. I removed the shared linking for now as requested.

PR #352, opened 2021-08-05
ci: run script to test bpftool types/options sync

qmonnet:
### Consistency checks for bpftool

This PR adds a call to test_bpftool_synctypes.py to run_selftests.sh. The objective of this script is to make sure that the different parts of bpftool that should be updated on BPF UAPI changes (type lists, help messages, documentation, bash completion) do remain in sync between them, and with the UAPI headers from the kernel. The motivation to have it in the CI is to effectively make sure that new patchsets consistently update all those parts when necessary.

This PR is similar to https://github.com/kernel-patches/vmtest/pull/22.

### DIFF
--- a/travis-ci/vmtest/run.sh
+++ b/travis-ci/vmtest/run.sh
@@ -386,6 +386,15 @@ LIBBPF_PATH="${REPO_ROOT}" \
 	REPO_PATH="${REPO_PATH}" \
 	VMLINUX_BTF=${vmlinux} ${VMTEST_ROOT}/build_selftests.sh
 
+travis_fold start bpftool_checks "Running bpftool checks..."
+if [[ "${KERNEL}" = 'LATEST' ]]; then
+	"${REPO_ROOT}/${REPO_PATH}/tools/testing/selftests/bpf/test_bpftool_synctypes.py" && \
+		echo "Consistency checks passed successfully."
+else
+	echo "Consistency checks skipped."
+fi
+travis_fold end bpftool_checks
+
 travis_fold start vm_init "Starting virtual machine..."
 
 if (( SKIPSOURCE )); then


thefallentree (approved):

thefallentree:
```  /usr/bin/env: 'python3': No such file or directory``` 
looks like we need python3

qmonnet:
Ok I updated the commit to install it in build_selftests.sh, let's see if it works better

anakryiko:
This PR reminds me that libbpf CI is a better place to start and test something like this, as Github Actions actually would test this based on not-yet-committed changes. With kernel-patches CI the setup is more complicated so testing ahead of time doesn't work.

anakryiko (changes requested):
doesn't seem to help...

qmonnet:
Yep I noticed, I'll look into this.

qmonnet:
I updated mkrootfs.sh but I can't push a new Arch image from there. Is it acceptable to add Python3 to the image, and if so, can someone help?

qmonnet:
Yes I know, I think it should work. I would have preferred to run it in the VM because 1) it looked cleaner to have all tests in it and 2) I've got another Python3 script in progress somewhere to test more things with bpftool and which would need to run in the VM if we ever considered it for inclusion (in a future where I can find the time to finish it). But I guess the discussion can wait until I have something ready to submit.

I can understand you prefer to keep the current image without Python3, I'll rework my PR to run the script outside of the VM.

anakryiko:
I don't mind adding Python to the image per se, but it's not the highest priority right now. I'll make a note to add this as another work item for BPF CI, but for now I'd say let's run this right after selftests (it also logically fits better).

qmonnet:
I updated the PR to run the checks on the host, after building the selftests. Looks like this is working this time.


latest

anakryiko:
Looks good, thank you. Please do the same for kernel-patches/bpf (if you haven't yet). Also, we started consolidating testing logic between kernel-patches and libbpf, so we now have a third repo (temporarily) at https://github.com/libbpf/ci, can you please duplicate this PR to that repo as well. I'm sorry that it's a bit of a pain right now, but we are working towards having all this consolidated.

PR #353, opened 2021-08-06
makefile: fix missing object for static compilation

rafaeldtinoco:
Makefile needs relo_core object added to objects list to avoid static
linking errors when doing static compilation:

/bin/ld: .../libbpf.a(libbpf.o): in function `bpf_core_apply_relo':
.../libbpf/src/libbpf.c:5134: undefined reference to `bpf_core_apply_relo_insn'

Signed-off-by: Rafael David Tinoco <rafaeldtinoco@gmail.com>


rafaeldtinoco:
I'm sending this here as this makefile is not the same as the one in kernel source.

rafaeldtinoco:
I see there is already on on going fix for this so I'm closing this.

anakryiko:
Thanks! This is a good fix, applying.

PR #366, opened 2021-08-18
Libbpf sync 2021-08-18T20:34:03.609Z

grantseltzer:
[Full output of script](https://gist.github.com/grantseltzer/97bdbc20a229d8605ad713f8968b65c6)


anakryiko:
We need to do something with the branch name to PR name conversion, I always manually fix it up so that it looks sane as UTC timestamp. Please do the same until we have something better.

grantseltzer:
Sounds good!

PR #382, opened 2021-09-16
Add enum to be displayed in documentation

grantseltzer:
This adds the enums in the header files to be displayed on libbpf.readthedocs.org

Follow up question to this - is everything in the headers a public facing API, and everything in c files private? Is that how things are separated?

Signed-off-by: grantseltzer <grantseltzer@gmail.com>


anakryiko:
Yep about public vs private. Some macro in public headers are not considered to be part of public API, if they start with underscores (it's just because you can't really hide macros in headers).

PR #387, opened 2021-10-01
Remove caching of pahole test result

tsipa:
Initial idea was to run it hourly, but when it runs hourly it produce a
lot of useless noise and we had to switch it do daily. When we run it
daily caching make much less sense and only make debugging more complex.


anakryiko (approved):
LGTM, thanks!

PR #392, opened 2021-10-20
[ci] skip forks

evverx:
[coverity] skip forks

to stop GitHub from sending notifications about failed attemtps
to run the Coverity workflow without `COVERITY_SCAN_TOKEN` like
https://github.com/evverx/libbpf-fork/actions/runs/1364759947

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   coverity:
     runs-on: ubuntu-latest
+    if: github.repository == 'libbpf/libbpf'
     name: Coverity
     steps:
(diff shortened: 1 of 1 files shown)


evverx:
> We haven't hit any issues with free Github Actions account on https://github.com/kernel-patches/bpf/pulls, and there we run heavy-weight runs much more frequently and for much more PRs all the time

I'm pretty sure that's probably because it's free and public. It would be a little different if it was a private repository.

> One can also argue that if you need to save some time you can add that if: line in your own fork, no? ;)

Fair enough :-)

I'll drop the second commit then

anakryiko:
thanks

PR #393, opened 2021-10-21
[ci] turn on CIFuzz

evverx:
Just to make sure it works

https://google.github.io/oss-fuzz/getting-started/continuous-integration/

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,40 @@
+---
+# https://google.github.io/oss-fuzz/getting-started/continuous-integration/
+name: CIFuzz
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


anakryiko:
I think it's useful to have for sure. I can't promise to fix all the fuzzed bugs, but we can open issues and work through the backlog to improve libbpf's resilience to corrupted inputs over time. It would be good for CIFuzz jobs to not mark build as failed (red), though. Do we need to mark the workflow as `continue-on-error: true` to achieve that? Other than that, I think we should merge it for sure.

evverx:
> Do we need to mark the workflow as continue-on-error: true to achieve that?

I don't think it would be necessary. CIFuzz is smart enough to automatically figure out whether PRs introduce new bugs or not by trying to reproduce crashes using the latest build it downloads from OSS-Fuzz. That's what happened in https://github.com/libbpf/libbpf/runs/3980664985. The fuzz target crashed but since the same testcase crashed the latest build from OSS-Fuzz CIFuzz decided not to report the crash and the workflow turned green.

PR #406, opened 2021-11-16
ci: account for bpftool check results, print a summary of all tests

qmonnet:
This is the libbpf port of the PR for the kernel CI, at https://github.com/kernel-patches/vmtest/pull/44.

Original PR description says:

> The bpftool checks work as expected when the CI runs, except that they do not set any error status code for the script on error, which means that the failures are lost among the logs and not reported in a clear way to the reviewers.
>
> Let's account for the status of bpftool checks when exiting from run.sh
>
> Let's also create a summary for the different group tests, at the end of the test logs, under the folded sections. This summary may be extended in the future with a more detailed overview of the tests running in the VM.

### DIFF
--- a/travis-ci/vmtest/run.sh
+++ b/travis-ci/vmtest/run.sh
@@ -410,10 +410,20 @@ LIBBPF_PATH="${REPO_ROOT}" \
 	REPO_PATH="${REPO_PATH}" \
 	VMLINUX_BTF=$(realpath ${source_vmlinux}) ${VMTEST_ROOT}/build_selftests.sh
 
+declare -A test_results
+
 travis_fold start bpftool_checks "Running bpftool checks..."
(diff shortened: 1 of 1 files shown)


qmonnet:
The bpftool checks reports inconsistencies that were [fixed upstream in bpf-next](https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/tools/bpf/bpftool?id=3811e2753a39efb8aa5b8c133dc24f6d26f6cd96), but that are beyond the commit referenced in `CHECKPOINT-COMMIT` and then not pulled in the GH Action - so the failure is expected.

We can rebase the PR after the next libbpf synchronisation, to make sure that all is in order.

anakryiko:
@qmonnet I've synced libbpf to latest, can you please rebase?

PR #414, opened 2021-12-01
ci: fix test on `/exitstatus` existence and size

qmonnet:
The condition for the test is incorrect, we want to add a default exit status if the file is empty. But [[ -s file ]] returns true if the file exists and has a size greater than zero. Let's reverse the condition.

[ Apologies for the silly mistake, previous PR was merged before I had time to check the result of the latest CI run :/ ]

Fixes: #412


qmonnet:
We didn't have all the tests appearing in the summary previously - Looks better now.

Screenshot_20211201_222318

qmonnet:
Although I wonder what happens to `test_maps`? :thinking:

PR #422, opened 2021-12-14
Libbpf sync 2021-12-14T22:25:46.870Z

anakryiko:

PR #424, opened 2021-12-16
ci: add LIRC kernel config

anakryiko:
Add necessary kernel config values to make BPF_PROG_TYPE_LIRC_MODE2
programs work.


thefallentree:
stack smashing on s390x ? do we need to run kernel tests on s390x to catch this earlier?

anakryiko:
> stack smashing on s390x ? do we need to run kernel tests on s390x to catch this earlier?

yep, we should think about running s390x for kernel-patches as well. For now I pined Ilya to help with s390x failure (https://github.com/libbpf/libbpf/pull/422#issuecomment-996027895)

PR #427, opened 2021-12-20
Add documentation for error checking in API doc

grantseltzer:
The API doc only exists in the github repo, so opened the PR here instead of the mailing list.

Signed-off-by: grantseltzer <grantseltzer@gmail.com>

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -6,7 +6,48 @@
 
 
 LIBBPF API
-==================
+==========
+
+Error Handling
+--------------
+
+When libbpf is used in "libbpf 1.0 mode", API functions can return errors in one of two ways.
+
+You can set "libbpf 1.0" mode with the following line:
+
+.. code-block::
+
+    libbpf_set_strict_mode(LIBBPF_STRICT_DIRECT_ERRS | LIBBPF_STRICT_CLEAN_PTRS);
+
+If the function returns an error code directly, it uses 0 to indicate success
+and a negative error code to indicate what caused the error. In this case the
+error code should be checked directly from the return, you do not need to check
+errno. 
+
+For example:
+
+.. code-block::
+
+    err = some_libbpf_api_with_error_return(...);
+    if (err < 0) {
+        /* Handle error accordingly */
+    }
+
+If the function returns a pointer, it will return NULL to indicate there was 
+an error. In this case errno should be checked for the error code.
+
+For example:
+
+.. code-block::
+
+    ptr = some_libbpf_api_returning_ptr();
+    if (!ptr) {
+        if (errno = -EINVAL) {
+           /* handle EINVAL error */
+        } else if (errno == E2BIG) { /* note no minus sign */
+           /* handle E2BIG error */
+        }
+    }
 
 libbpf.h
 --------


docs/api.rst
```diff
@@ -6,7 +6,38 @@
...
+and a negative error code to indicate what caused the error. In this case the
+error code should be checked directly from the return, you do not need to check
+errno. 
+
+For example:
+
+```
+    err = some_libbpf_api_with_error_return(...);
+    if (err < 0) {
+        // Handle error accordingly
```
anakryiko, line 26: use `/* */` :-P

docs/api.rst
```diff
@@ -6,7 +6,38 @@
...
+
+If the function returns a pointer, it will return NULL to indicate there was 
+an error. In this case errno should be checked for the error code.
+
+For example:
+
+```
+    ptr = some_libbpf_api_returning_ptr(...);
+    if (!ptr) {
+       /* errno has error */
```
anakryiko, line 38: I guess add a code like `if (errno == EINVAL) { /* handle error */ }` to make it unambiguous what "errno" we refer to?

docs/api.rst
```diff
@@ -6,7 +6,38 @@
 
 
 LIBBPF API
-==================
+==========
```
anakryiko, line 9: Please remind me, what did we agree about with where libbpf docs live? Not in linux repo anymore, right? I see that we have libbpf_build.rst, libbpf_naming_convention.rst, index.rst that are copied from Linux repo. But api.rst is only in libbpf/libbpf on Github? If we moved everything to github, why didn't we move existing docs here as well?

docs/api.rst
```diff
@@ -6,7 +6,38 @@
...
+
+Error Handling
+--------------
+
+API functions in libbpf can return errors in one of two ways.
```
anakryiko, line 14: Can you please add a note that all this applies to "libbpf 1.0" mode which can be set with `libbpf_set_strict_mode(LIBBPF_STRICT_DIRECT_ERRS | LIBBPF_STRICT_CLEAN_PTRS);`?

docs/api.rst
grantseltzer: So as it stands now the libbpf documentation (the files you listed) live in the kernel repo, except for the API documentation (specifically `api.rst`) which only lives in the mirror. This is because the API documentation should be versioned and having the 'latest' API docs in the kernel repo, and therefore on kernel.org/doc is ambiguous and confusing.

The configuration files for sphinx/doxygen/readthedocs (what powers libbpf.readthedocs.org) also live in the mirror.

grantseltzer: That's why index.rst links to libbpf.readthedocs.org explicitly, not a relative link.

docs/api.rst
grantseltzer: Can you elaborate on how this works? Will this be default starting at 1.0? Is it not on by default now?

grantseltzer:
Also just to note, you can see a preview of these changes here: 

https://libbpf-test.readthedocs.io/en/latest/api.html

(Still need to get to having a makefile target to do that locally, I promise I will soon!)

docs/api.rst
anakryiko: This will be the only behavior in 1.0, yes. It's not on by default right now, you have to "opt-in" for the transition period before 1.0

docs/api.rst
```diff
@@ -6,7 +6,49 @@
 
 
 LIBBPF API
-==================
+==========
+
+Error Handling
+--------------
+
+When libbpf is used in "libbpf 1.0 mode",API functions can return errors in one of two ways.
```
anakryiko, line 14: space missing after comma

docs/api.rst
```diff
@@ -6,7 +6,49 @@
...
+and a negative error code to indicate what caused the error. In this case the
+error code should be checked directly from the return, you do not need to check
+errno. 
+
+For example:
+
+```
+    err = some_libbpf_api_with_error_return(...);
+    if (err < 0) {
+        /* Handle error accordingly */
```
anakryiko, line 32: hm... I think this example doesn't show that both err and errno can be used, let's add this snippet:

```
...
if (err < 0) {
    if (err = -EINVAL) {
        /* handle EINVAL error */
    } else if (errno == E2BIG) { /* note no minus sign */
        /* handle E2BIG error */
    }
```

This should make clear that both err and errno have error code set (and hopefully people will notice that errno has positive error code, while err is negative)

docs/api.rst
anakryiko: something is wrong with rendering, btw 
Screen Shot 2021-12-22 at 1 42 30 PM

anakryiko (changes requested):

docs/api.rst
grantseltzer: Should be fixed now.

docs/api.rst
```diff
@@ -6,7 +6,48 @@
...
+    }
+
+If the function returns a pointer, it will return NULL to indicate there was 
+an error. In this case errno should be checked for the error code.
+
+For example:
+
+.. code-block::
+
+    if (err < 0) {
```
anakryiko, line 43: I think you mixed up examples. For pointer returning APIs you don't have `err` anymore and can only use errno. While the example above could show that err and errno could be used almost interchangeably (except errno and err sign differs and errno can be overriden by any glibc or libbpf API function)

grantseltzer: Ah my mistake, fixed that. 

In the case of non-pointer API functions we could document that errno and the returned code could be used interchangeably. I don't think we should though, better to keep it simple. wdyt?

docs/api.rst
```diff
@@ -6,7 +6,48 @@
...
+If the function returns a pointer, it will return NULL to indicate there was 
+an error. In this case errno should be checked for the error code.
+
+For example:
+
+.. code-block::
+
+    ptr = some_libbpf_api_returning_ptr();
+    if (!ptr) {
+        if (errno = -EINVAL) {
```
anakryiko, line 45: this shouldn't have minus sign either if we are using errno. I'll fix it up when I apply it.

grantseltzer: also missed the `==`, sorry about that. I shouldn't commit when very tired.

PR #428, opened 2021-12-20
Adding a step to patch kernel

thefallentree:


thefallentree:
@anakryiko  can you rerun those build jobs,  docker hub is having a moment just now

Issue #455, opened 2022-02-11
Incorrect version exported to libbpf.pc


michel-slm:
This commit: https://github.com/libbpf/libbpf/commit/8ade99a6f84d0d7e70965451533b8d76926ee391 stops the libbpf version from being dynamically generated; instead it is sourced from libbpf.map. But libbpf.map only lists 0.x.0 versions - since presumably the ABI is unchanged between patch releases.

Unfortunately this then leaks to the Makefile, which means it leaks to libbpf.pc, which means libbpf-0.6.1 claims to be 0.6.0 and any app depending on 0.6.1 fails to find its dependencies.

Proposed workaround in Fedora (passing LIBBPF_VERSION to make): https://src.fedoraproject.org/rpms/libbpf/pull-request/2

PR #459, opened 2022-02-20
libbpf: fix libbpf.pc generation w.r.t. patch versions

anakryiko:
Ensure that libbpf.pc gets full libbpf's version, including patch
releases. Also add some mechanism to ensure that official released
version (e.g., 0.7.1) and the one recorded in libbpf.map (which never
bumps patch version, so will be 0.7.0) are in sync up to major and minor
versions. This should ensure that major mistakes are captured. We'll
still need to be very careful with zeroing out patch version on minor
version bumps.

Closes: https://github.com/libbpf/libbpf/issues/455
Reported-by: Michel Salim <michel@fb.com>

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -8,10 +8,15 @@ else
 	msg = @printf '  %-8s %s%s\n' "$(1)" "$(2)" "$(if $(3), $(3))";
 endif
 
-LIBBPF_VERSION := $(shell \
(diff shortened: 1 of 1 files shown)


anakryiko:
@michel-slm, would something like this work?

michel-slm:
Sorry for commenting late @anakryiko but yeah, this should work. I like the failsafe to check if the two versions are mismatched as well.

PR #463, opened 2022-02-26
mirror: Ship linux/fcntl.h linux/openat2.h UAPI header

jameshilliard:
Copy tools/include/uapi/linux/fcntl.h from kernel's repository, taken at commit 75134f16e7dd ("bpf: Add schedule points in batch ops") (current checkpoint).

We want this header in the repo so we can be sure that it contains the flags `F_DUPFD_CLOEXEC`, used in libbpf_internal.h.

We also need to include `$(TOPDIR)/include/uapi/linux` so that the right `fcntl.h` header gets used.

Fixes:
```c
libbpf_internal.h:521:18: error: 'F_DUPFD_CLOEXEC' undeclared (first use in this function); did you mean 'FD_CLOEXEC'?
   fd = fcntl(fd, F_DUPFD_CLOEXEC, 3);
                  ^~~~~~~~~~~~~~~
                  FD_CLOEXEC
```


jameshilliard:
Maybe just a conditional define like in #464 is better.

PR #471, opened 2022-03-27
[BUG] fix tc flag

Super-long:
I encountered a problem when using tail call in TC, my main program and tail call BPF program are named like "tc/prog_a" and "tc/prog_b", but encountered such a problem "missing BPF prog type, check ELF section name", after checking the code I found that the problem lies in "sec_def_matches", for such a name of "tc/prog_b" will not match any rules in "sec_def_matches", it will cause "find_sec_def" to fail and cause prog ->type assignment failed.

Also "classifier" is currently declared as "SEC_DEPRECATED", which contains "SEC_SLOPPY_PFX", and the bpf_prog_type of "classifier" and "tc" are both SCHED_CLS, I think such vague prefix matching requirements are reasonable, so I think it may be is a bug. And when I add "SEC_SLOPPY_PFX" to "tc" my program can execute normally.


chenhengqi:
For such use cases, you can use custom SEC() handler:

https://github.com/libbpf/libbpf/blob/3591deb9bc6b0848721f831790b78ae0a593b4c5/src/libbpf.h#L1385-L1438

anakryiko:
SEC("tc") is strict by design and intentionally. You shouldn't need to use tc/prog_a and tc/prog_b, both programs should be marked with SEC("tc"). Section name is not a unique identifier for programs.

PR #492, opened 2022-05-09
vmtest: use debian bookworm

iii-i:
Hi,

This is the fix for https://lore.kernel.org/bpf/CAEf4BzbqfV1FySVu69-am6hz2ykPWXspqyFV8eG_Gapwp0kUqw@mail.gmail.com/.
You can build a new image yourself or take mine from https://github.com/iii-i/libbpf/raw/libbpf-vmtest-rootfs-2022.05.09/travis-ci/rootfs/libbpf-vmtest-rootfs-2022.05.09.tar.zst.
I tested it with `[PATCH bpf-next v3 0/8] bpf: mptcp: Support for mptcp_sock and is_mptcp` using vmtest.sh.

Best regards,
Ilya

### DIFF
--- a/travis-ci/rootfs/mkrootfs_debian.sh
+++ b/travis-ci/rootfs/mkrootfs_debian.sh
@@ -16,7 +16,7 @@ trap 'rm -r "$root"' EXIT
 
 # Install packages.
 packages=binutils,busybox,elfutils,ethtool,iproute2,libcap2,libelf1,strace,zlib1g
-debootstrap --include="$packages" --variant=minbase bullseye "$root"
+debootstrap --include="$packages" --variant=minbase bookworm "$root"
 
 # Remove the init scripts (tests use their own). Also remove various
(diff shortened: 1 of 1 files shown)


anakryiko:
thanks, @iii-i! @mykola-lysenko, should we just upload Ilya's latest image or it's better to regenerate?

chantra:
> thanks, @iii-i! @mykola-lysenko, should we just upload Ilya's latest image or it's better to regenerate?

We should just upload that one as we don't (or at least I) have a s390x handy to create the image and AFAIK, we can only produce the images from a host in the same arch.

chantra:
> AFAIK, we can only produce the images from a host in the same arch.

I suppose this is not true since this very commit :) https://github.com/libbpf/libbpf/pull/492/commits/f30d1f7258d123d2451d10dfca36ee7458eac6af

PR #495, opened 2022-05-13
Makefile: add support for cross compilation

setNull:
Hello team,

I'd like to build bpftool and libbpf-bootstrap project for arm64 Linux, but it seems the support is not as well as host build.
I add some changes in makefile to make it. 
This one is the first one since libbpf is submodule of other projects.

Please let me know if this is the correct way to achieve that, thanks.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -18,6 +18,15 @@ ifneq ($(LIBBPF_MAJMIN_VERSION), $(LIBBPF_MAP_VERSION))
 $(error Libbpf release ($(LIBBPF_VERSION)) and map ($(LIBBPF_MAP_VERSION)) versions are out of sync!)
 endif
 
+define allow-override
+  $(if $(or $(findstring environment,$(origin $(1))),\
+            $(findstring command line,$(origin $(1)))),,\
+    $(eval $(1) = $(2)))
+endef
+
+$(call allow-override,CC,$(CROSS_COMPILE)cc)
+$(call allow-override,LD,$(CROSS_COMPILE)ld)
+
 TOPDIR = ..
 
 INCLUDES := -I. -I$(TOPDIR)/include -I$(TOPDIR)/include/uapi
@@ -26,8 +35,9 @@ ALL_CFLAGS := $(INCLUDES)
 SHARED_CFLAGS += -fPIC -fvisibility=hidden -DSHARED
 
 CFLAGS ?= -g -O2 -Werror -Wall -std=gnu89
-ALL_CFLAGS += $(CFLAGS) -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64
-ALL_LDFLAGS += $(LDFLAGS)
+ALL_CFLAGS += $(CFLAGS) -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 $(EXTRA_CFLAGS)
+ALL_LDFLAGS += $(LDFLAGS) $(EXTRA_LDFLAGS)
+
 ifdef NO_PKG_CONFIG
 	ALL_LDFLAGS += -lelf -lz
 else


jameshilliard:
Not entirely sure this is needed, we've been able to cross compile libbpf like [this](https://github.com/buildroot/buildroot/blob/master/package/libbpf/libbpf.mk).

src/Makefile
```diff
@@ -28,6 +31,12 @@ SHARED_CFLAGS += -fPIC -fvisibility=hidden -DSHARED
 CFLAGS ?= -g -O2 -Werror -Wall -std=gnu89
 ALL_CFLAGS += $(CFLAGS) -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64
 ALL_LDFLAGS += $(LDFLAGS)
+ifneq ($(EXTRA_CFLAGS),)
```
anakryiko, line 34: in general having EXTRA_CFLAGS is probably useful on its own, but there is no need for `ifneq`, you can just use it above when appending to ALL_CFLAGS, same for EXTRA_LDFLAGS

src/Makefile
```diff
@@ -18,6 +18,9 @@ ifneq ($(LIBBPF_MAJMIN_VERSION), $(LIBBPF_MAP_VERSION))
 $(error Libbpf release ($(LIBBPF_VERSION)) and map ($(LIBBPF_MAP_VERSION)) versions are out of sync!)
 endif
 
+$(if $(findstring default,$(origin CC)),$(eval CC = $(CROSS_COMPILE)$(CC)))
+$(if $(findstring default,$(origin LD)),$(eval LD = $(CROSS_COMPILE)$(LD)))
```
anakryiko, line 22: can't user just specify CC and LD like this:

```
make CC=<whatever> LD=<whatever>
```

?

setNull: > 

Yes, you can do that when building standalone.
However, user (like me) may want to build libbpf inside other tools or applications.
F.ex., 
when building inside bpftool, it supports CROSS_COMPILE env.
User builds like:
`make CROSS_COMPILE=xxx ARCH=xxx`,  will generate an arm64 bpftool trying to link a x86 libbpf.

That's why I add this change.

src/Makefile
setNull: Yeah, I just keep the same style as https://github.com/libbpf/bpftool/blob/master/src/Makefile#L83

setNull:
> Not entirely sure this is needed, we've been able to cross compile libbpf like [this](https://github.com/buildroot/buildroot/blob/master/package/libbpf/libbpf.mk).

Here I'd like to make it easier for building inside other tools like bpftool.

src/Makefile
ThinkerYzu1: I just checked the makefile of bpftool.  It passes HOSTCC as the value of CC to build libbpf for bootstrap. But, it doesn't give a value of CC to create a target binary of libbpf.  You could fix it at the bpftool side by providing a correct CC value.

By the definitions of BUILD, HOST, and TARGET machines, as I see it, bpftool build itself for BUILD and HOST machines while HOST equals TARGET for this case.  It uses the bpftool of the BUILD machine to generate header files at build time—however, the makefile of bpftool calls bootstrap HOST, which is confusing.

ThinkerYzu1:
IMHO, it is the responsibility of the project using libbpf to assign a correct CC.  In terms of cross-compiling,  we always build libbpf for the target machines.  It doesn't build any code for the build machine.

setNull:
> IMHO, it is the responsibility of the project using libbpf to assign a correct CC. In terms of cross-compiling, we always build libbpf for the target machines. It doesn't build any code for the build machine.

@ThinkerYzu1 hi, I just want to make it more convenient in case someone else need this.

ThinkerYzu1:
@setNull I have a feeling that you are trying to fix a bug in the makefiles of bpftool by changing the makefile here.  How about trying to fix bpftool instead of libbpf?

setNull:
> @setNull I have a feeling that you are trying to fix a bug in the makefiles of bpftool by changing the makefile here. How about trying to fix bpftool instead of libbpf?

I agree change either side (bpftool or libbpf) is OK.
However, not only bpftool, some other applications like bcc/libbpf-tools also have this as submodule, so I think it's better to enhance here.

ThinkerYzu1:
@setNull would you like to send a PR to bpftool as well?

setNull:
> @setNull would you like to send a PR to bpftool as well?

OK, this one works for me,  https://github.com/libbpf/bpftool/pull/25/files

But, I still hope this PR of libbpf can be accepted, to the benefit of easily building other tools like bcc/libbpf-tools , not only bpftool.

setNull:
@ThinkerYzu1 I sent a PR( https://github.com/libbpf/bpftool/pull/25) to bpftool, and got more information about these two repos.

Makefile of **bpftool** is keep as close as kernel repo's, while makefile of **libbpf** is rewritten from scratch.
In kernel repo, both tools/bpftool and tools/lib/bpf take the CROSS_COMPILE env(https://github.com/torvalds/linux/blob/master/tools/lib/bpf/Makefile#L52), so building for target with below CMD works well, so no change need to be sent to kernel repo.

`make CROSS_COMPILE=xxx CFLAGS=XXX LDFLAGS=XXX`

However, for makefile of standalone libbpf, it doesn't take the CROSS_COMPILE env, so above CMD not work.
Sounds like here is the right place to enhance. :-)

src/Makefile
anakryiko: Let's follow what tools/scripts/Makefile.include is doing and let's define allow-override function and call it. Keep it similar to how kernel is handling this?

src/Makefile
anakryiko: I'd change it in bpftool's Makefile as well. Please don't add extra lines for something as simple as this. Please inline `$(EXTRA_CFLAGS)` and `$(EXTRA_LDFLAGS)` in ALL_CFLAGS and ALL_LDFLAGS definitions

anakryiko (changes requested):
Ok, let's do this, but please address the feedback. Also, please add your Signed-off-by: tag with full name and email, just like for kernel patches. Libbpf follows that style in its Github mirror as well. Thanks!

src/Makefile
```diff
@@ -18,6 +18,15 @@ ifneq ($(LIBBPF_MAJMIN_VERSION), $(LIBBPF_MAP_VERSION))
 $(error Libbpf release ($(LIBBPF_VERSION)) and map ($(LIBBPF_MAP_VERSION)) versions are out of sync!)
 endif
 
+define allow-override
+  $(if $(or $(findstring environment,$(origin $(1))),\
+            $(findstring command line,$(origin $(1)))),,\
+    $(eval $(1) = $(2)))
+endef
+
+$(call allow-override,CC,$(CROSS_COMPILE)gcc)
```
anakryiko, line 27: can you please make it default to cc? That's what make uses as the default.

anakryiko (changes requested):
> Signed-off-by: wangjie <wangjie22@lixiang.com>

Can you please use your full and properly capitalized name?

anakryiko (approved):

PR #501, opened 2022-05-19
synproxy fixes

iii-i:
Hi,

These are the fixes for https://lore.kernel.org/bpf/20220503171437.666326-1-maximmi@nvidia.com/

I still couldn't get it to fully work - it now fails with `test_synproxy:FAIL:initial SYNACKs unexpected initial SYNACKs: got FALSE`, but at least iptables commands now succeed.
@nvmmax, could you have another look please?

The new image is here: https://github.com/iii-i/libbpf/raw/0e8ebd35de660a98a7d4d8e919670cac9e910bb5/libbpf-vmtest-rootfs-2022.05.19.tar.zst

Best regards,
Ilya


iii-i:
I've taken a closer look, and turns out this is caused by `xdp_synproxy` failing to load the BPF code with `JIT does not support calling kernel function`. This is one of the current limitation of s390x JIT; I'm afraid we'll have to blacklist the test for now.

nvmmax:
I see... Thanks for your analysis! That means all tests that call kfuncs must be blacklisted on s390x, right? Where should I blacklist my test: is it somewhere in the kernel makefiles, or in some CI repository?

iii-i:
I would add it here: https://github.com/libbpf/libbpf/blob/master/travis-ci/vmtest/configs/blacklist/BLACKLIST-latest.s390x

nvmmax:
Added:

https://github.com/libbpf/libbpf/pull/502
https://github.com/kernel-patches/vmtest/pull/84

Thanks!

mykola-lysenko:
@iii-i, we want to merge the travis-ci/rootfs/mkrootfs_debian.sh script, but I am not sure about all the changes in the travis-ci/vmtest/configs/config-latest.s390x. Can you update the PR to remove the config change? Or is there anything required in it?

iii-i:
I'd keep the config changes, since they are required by the iptables invocations in the synproxy test.

PR #530, opened 2022-07-08
Round up map->def.max_entries when reusing a ringbuf

telbizov:
Libbpf (tested on version 0.8.0 as well as HEAD built from source) has a bug were reusing an already existing pinned RINGBUF map results in a "parameter mismatch" due to incorrectly comparing the max_entries values of the existing (pinned) map and the one under consideration.

The issue seems to come from the fact that the pinned map already has its max_entries rounded up while the map under consideration doesn't. Details and a simple patch follow.

Map definition:

```
struct {
 __uint(type, BPF_MAP_TYPE_RINGBUF);
 __uint(max_entries, 5000);
 __uint(pinning, LIBBPF_PIN_BY_NAME);
} evring0 SEC(".maps");
```

First load (no map loaded):

```
578: ringbuf name evring0 flags 0x0
 key 0B value 0B max_entries 8192 memlock 24576B
```

Second load with my added debug output:

```
libbpf: Rumen Telbizov Debug: map_info.type (27) == map->def.type (27)
libbpf: Rumen Telbizov Debug: map_info.max_entries (8192) == map->def.max_entries (5000)
libbpf: couldn't reuse pinned map at '/sys/fs/bpf/evring0': parameter mismatch
libbpf: map 'evring0': error reusing pinned map
libbpf: map 'evring0': failed to create: Invalid argument(-22)
```

The issue lies in map_is_reuse_compat() which does the comparison. The following fixes the issue:

```
- map_info.max_entries == map->def.max_entries &&
+ map_info.max_entries == (map_info.type == BPF_MAP_TYPE_RINGBUF ? adjust_ringbuf_sz(map->def.max_entries) : map->def.max_entries) &&
```

### DIFF
--- a/src/libbpf.c
+++ b/src/libbpf.c
@@ -4750,6 +4750,37 @@ bool kernel_supports(const struct bpf_object *obj, enum kern_feature_id feat_id)
 	return READ_ONCE(feat->res) == FEAT_SUPPORTED;
 }
 
+static size_t adjust_ringbuf_sz(size_t sz)
+{
+	__u32 page_sz = sysconf(_SC_PAGE_SIZE);
+	__u32 mul;
+
+	/* if user forgot to set any size, make sure they see error */
+	if (sz == 0)
+		return 0;
+	/* Kernel expects BPF_MAP_TYPE_RINGBUF's max_entries to be
+	 * a power-of-2 multiple of kernel's page size. If user diligently
+	 * satisified these conditions, pass the size through.
+	 */
+	if ((sz % page_sz) == 0 && is_pow_of_2(sz / page_sz))
+		return sz;
+
+	/* Otherwise find closest (page_sz * power_of_2) product bigger than
+	 * user-set size to satisfy both user size request and kernel
+	 * requirements and substitute correct max_entries for map creation.
+	 */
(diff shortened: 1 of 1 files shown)


anakryiko:
The problem makes sense, but this has to be fixed upstream in kernel repo. So please submit the patch to bpf@vger.kernel.org. 

I also think that it will be cleaner to eagerly adjust max_entries as soon as possible, which currently seems to be two places: in fill_map_from_def() and in bpf_map__set_max_entries(). So consider doing that instead, that way max_entries will always stay correct and adjusted.

telbizov:
@anakryiko I am not very familiar with the code base and I should have made it clear that this was just a proof of concept patch. Would you be OK if I leave it with you to patch the best way you find and in the appropriate upstream repository?

Thanks

anakryiko:
Submitted as https://patchwork.kernel.org/project/netdevbpf/list/?series=660198&state=*

telbizov:
Much appreciated. Feel free to close this PR and the associated issue whenever you deem appropriate.

anakryiko:
It will be auto-closed when the fix is synced from upstream kernel repo to Github.

Issue #529, opened 2022-07-08
Backport bugfix from xdp-project/xdp-tools#205


alexforster:
Would y'all accept a PR to backport the bugfix in xdp-project/xdp-tools#205 to a new libbpf branch? I see XDP support was removed in master recently, but for various reasons it's going to be a while before we move to libxdp, so I'd appreciate it if we could cut a `v0.8.1` release with this fix.

anakryiko:
Can you send PR against https://github.com/libbpf/libbpf/tree/v0.8.0 ? I can cut 0.8.1 with that.

anakryiko:
@alexforster created https://github.com/libbpf/libbpf/tree/libbpf-0.8.1

anakryiko:
https://github.com/libbpf/libbpf/releases/tag/v0.8.1

alexforster:
Thanks!

PR #533, opened 2022-07-14
libbpf: make shared xsk creation network namespace aware

Dhiver:
Creating AF_XDP sockets with the XDP_SHARED_UMEM flag set fails when two devices
ifindex (and the queue_id) are the same. This can happen with devices in
different network namespaces.

In the shared UMEM mode, each Xsk is expected to have a dedicated fill and
completion queue. Context data about those queues is stored in a linked-list
within the UMEM object.
The comparison performed to pick the right context for an Xsk prior to calling
bind() only takes into account the ifindex and queue_id of the network device.
However, the ifindex is not a "system-wide" unique identifier, it is unique only
within a network namespace. This results in attributing the same xsk_ctx to
different Xsks.

To fix this, we retrieve the netns_cookie associated with the socket at its
creation and add it in the comparison operation. This value "provides a global
identifier that can be assumed unique" ([0]).
We use the netns cookie instead of the netns inode number because inode numbers
can be reused.

  [0]: https://lore.kernel.org/bpf/c47d2346982693a9cf9da0e12690453aded4c788.1585323121.git.daniel@iogearbox.net/

Signed-off-by: Bastien Dhiver <bastien@cloudflare.com>

### DIFF
--- a/src/xsk.c
+++ b/src/xsk.c
@@ -52,6 +52,12 @@
  #define PF_XDP AF_XDP
 #endif
 
+#ifndef SO_NETNS_COOKIE
+ #define SO_NETNS_COOKIE 71
+#endif
+
+#define INIT_NS 1
+
 enum xsk_prog {
 	XSK_PROG_FALLBACK,
 	XSK_PROG_REDIRECT_FLAGS,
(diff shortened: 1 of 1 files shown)


Dhiver:
See: #529

Dhiver:
Can someone help me understand the CI errors and how they are related to the proposed change please?
image

anakryiko:
@Dhiver they are unrelated to your PR, I think.

But cc @danielofcb just in case, I might have forgotten, but I thought we haven't removed those from blacklist yet?

PR #535, opened 2022-07-19
fix resolve_full_path; .so does not necessarily have X permission

gfx:
SSIA. Otherwise `SEC("libc.so.6:...")` does not work. We can avoid this issue by specifying the full path of the executable or shared object instead of its basename, though.

### DIFF
--- a/src/libbpf.c
+++ b/src/libbpf.c
@@ -10466,8 +10466,12 @@ static int resolve_full_path(const char *file, char *result, size_t result_sz)
 {
 	const char *search_paths[3] = {};
 	int i;
+	bool is_executable = !(str_has_sfx(file, ".so") || strstr(file, ".so."));
+	int required_perms = R_OK;
+	if (is_executable)
+		required_perms |= X_OK;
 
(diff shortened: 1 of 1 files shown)


chenhengqi:
Patch like this should be sent to BPF mailing list.

anakryiko:
@gfx are you planning to submit upstream patch for this?

gfx:
@anakryiko Apologies, but I have no time to do so right now. Hope the project accepts pull-requests via GitHub.

chenhengqi:
I can take this if you don't mind.

anakryiko:
@chenhengqi sure, please do, let's get all such small fixes and improvements into the libbpf 1.0. I'd like to release it officially some time soon. Thanks!

chenhengqi:
OK, will do.

chenhengqi:
Addressed by https://github.com/libbpf/libbpf/pull/546/commits/01a23f1f7200ea0e96719c37bd29f5760573eb93.

gfx:
Thank you, @chenhengqi!

PR #536, opened 2022-07-22
Preserve errno when calling libbpf_print().

saimizi:
pr_*() functions may be called from a context that errno need to be
preserved.

For example, In perf_event_open_probe(), pr_warn() is called before
return an error. And at caller side, bpf_program__attach_kprobe_opts()
check the errno value to report an error.

if pr_warn() clear the errno, some strange issue will happen like
follow:

Attaching 1 Kprobes.
WARN  : libbpf: kprobe perf_event_open() failed: Invalid argument
WARN  : libbpf: prog 'stacktrace_kb': failed to create kprobe 'vfs_caches_init+0x0' perf event: Success

First WARN is from perf_event_open_probe(), error is "Invalid argument"
and the second WARN is from bpf_program__attach_kprobe_opts() which
faied to recognaize the error and report as success.

Because the log function libbpf_print() used in  pr_*() is in fact a
callback function from the user, there is no guarantee for errno to be
preserved in it.

This commit saved and restored errno when calling libbpf_print().

Signed-off-by: Seimizu Joukan <joukan.seimizu@gmail.com>

### DIFF
--- a/src/libbpf_internal.h
+++ b/src/libbpf_internal.h
@@ -148,7 +148,9 @@ extern void libbpf_print(enum libbpf_print_level level,
 
 #define __pr(level, fmt, ...)	\
 do {				\
+	int __saved_errno = errno;							\
 	libbpf_print(level, "libbpf: " fmt, ##__VA_ARGS__);	\
+	errno = __saved_errno;								\
 } while (0)
 
 #define pr_warn(fmt, ...)	__pr(LIBBPF_WARN, fmt, ##__VA_ARGS__)


anakryiko:
Libbpf code generally saves error before calling pr_xxx(), but adding a bit more convenience this way might not be bad. But this should come through bpf-next Linux repo, where libbpf source code lives. Please consider submitting patch.

saimizi:
It seems that calling set_print() from libbpf-rs will clear errno which breaks the state of the Link created by program attach. Sure, I will consider to submit the patch to bpf-next later. 

Thank you.

anakryiko:
cc @insearchoflosttime do we need some fix on libbpf-rs side for this?

insearchoflosttime:
Yes, we'll need to preserve the errno in set_print(). I'll file an issue for this and get it fixed

anakryiko:
FYI, I sent https://patchwork.kernel.org/project/netdevbpf/patch/20220810183425.1998735-1-andrii@kernel.org/ to address this on libbpf side

chenhengqi:
Addressed by https://github.com/libbpf/libbpf/pull/546/commits/9b713c2f7f4ec41704a4f9af026b5f6d441cb0b3.

PR #537, opened 2022-07-28
Rely on upstream kernel configuration

danielocfb:
So far we have relied on the kernel configuration as checked into the
this repository. However, a suitable configuration is now included in
upstream Linux [0].
With this change we add support for using the configuration from there.

[0] https://lore.kernel.org/bpf/165893461358.29339.11641967418379627671.git-patchwork-notify@kernel.org/T/#m2a97b0ea9ef0ddee7a53bbf7919e3f324b233937

### DIFF
--- a/.github/actions/vmtest/action.yml
+++ b/.github/actions/vmtest/action.yml
@@ -43,7 +43,10 @@ runs:
       run: |
         echo "::group::Prepare buidling selftest"
         cd .kernel
-        cp ${{ github.workspace }}/travis-ci/vmtest/configs/config-latest.${{ inputs.arch }} .config
+        cat tools/testing/selftests/bpf/config \
+            tools/testing/selftests/bpf/config.${{ inputs.arch }} \
+            ${{ github.workspace }}/travis-ci/vmtest/configs/config-latest \
+            ${{ github.workspace }}/travis-ci/vmtest/configs/config-latest.${{ inputs.arch }} 2> /dev/null > .config && :
         make olddefconfig && make prepare
         cd -
         echo "::endgroup::"


danielocfb:
Will have to bump bpf-next CLN we sync to first.

mykola-lysenko:
for my own understanding: we will use configuration from upstream and blend it with local configuration. In case of 4.9.0 and 5.5.0 kernel version, we would not have upstream configuration available, right?

.github/actions/vmtest/action.yml
```diff
@@ -43,7 +43,10 @@ runs:
       run: |
         echo "::group::Prepare buidling selftest"
         cd .kernel
-        cp ${{ github.workspace }}/travis-ci/vmtest/configs/config-latest.${{ inputs.arch }} .config
+        cat tools/testing/selftests/bpf/config \
+            tools/testing/selftests/bpf/config.${{ inputs.arch }} \
+            ${{ github.workspace }}/travis-ci/vmtest/configs/config-latest \
+            ${{ github.workspace }}/travis-ci/vmtest/configs/config-latest.${{ inputs.arch }} 2> /dev/null > .config && :
```
mykola-lysenko, line 49: do we need :  here to ensure cat always return true?

danielocfb: Yep, that is my understanding.

danielocfb:
> for my own understanding: we will use configuration from upstream and blend it with local configuration. In case of 4.9.0 and 5.5.0 kernel version, we would not have upstream configuration available, right?

Correct, in case of 4.9 and 5.5 we would not have the configuration available upstream and only use the local one.

mykola-lysenko (approved):

PR #543, opened 2022-08-09
Remove checked-in configuration

danielocfb:
Both the bpf and bpf-next tree have suitable BPF selftest configurations
available for usage with the latest kernel now upstream. While we do
test on 4.9 and 5.5 kernels as well, there we just download prebuilt
binaries. The configuration we use for building selftests is always the
upstream one.
With this change we remove the checked-in configuration, as it is now no
longer needed.


danielocfb:
This pull request is kind of a follow up to https://github.com/libbpf/libbpf/pull/537. I wanted to land that one first to make sure that no issues occur, but conceptually they could have been a single change as well.

.github/actions/vmtest/action.yml
```diff
@@ -41,12 +41,10 @@ runs:
     - name: Prepare to build BPF selftests
       shell: bash
       run: |
-        echo "::group::Prepare buidling selftest"
+        echo "::group::Prepare buildling selftest"
```
mykola-lysenko, line 44: building?

.github/actions/vmtest/action.yml
```diff
@@ -41,12 +41,10 @@ runs:
...
         cd .kernel
         cat tools/testing/selftests/bpf/config \
-            tools/testing/selftests/bpf/config.${{ inputs.arch }} \
-            ${{ github.workspace }}/travis-ci/vmtest/configs/config-latest \
-            ${{ github.workspace }}/travis-ci/vmtest/configs/config-latest.${{ inputs.arch }} 2> /dev/null > .config && :
+            tools/testing/selftests/bpf/config.${{ inputs.arch }} > .config
         make olddefconfig && make prepare
         cd -
         echo "::endgroup::"
```
mykola-lysenko, line 50: not related to this change. can we reuse helpers to start / end group in here?

.github/actions/vmtest/action.yml
danielocfb: Oh yeah, not sure what happened here.

.github/actions/vmtest/action.yml
danielocfb: I believe we can, yes. I'll do it as a follow up change.

danielocfb:
I've addressed the review comment.

mykola-lysenko (approved):

PR #545, opened 2022-08-09
Rename blacklists and whitelists

danielocfb:
Upstream uses denylist and allowlist terminology instead of blacklist
and whitelist. It also has established a less deeply nested directory
structure.
This change renames the blacklist & whitelist files accordingly and
moves them one level up out of their containing directory to mirror the
layout we have upstream as well as in kernel-patches/vmtest.


mykola-lysenko (approved):

danielocfb:
Will merge this change tomorrow. Thanks for the review, Mykola!

PR #547, opened 2022-08-15
Re-enable test_maps selftest

danielocfb:
Back in 2020, we disabled the test_maps selftest with e05f9be4f4be
("vmtests: temporarily disable test_maps"), for reasons not closely
elaborated.
It appears that by now the test is succeeding again, so let's enable it
back.


mykola-lysenko:
Wow good find! Thanks for enabling it back

mykola-lysenko (approved):

PR #548, opened 2022-08-16
Topic/run vmtests improvements

danielocfb:
Please see individual commits for descriptions. Only grouped together because of GitHub limitations.

### DIFF
--- a/travis-ci/vmtest/run_selftests.sh
+++ b/travis-ci/vmtest/run_selftests.sh
@@ -26,7 +26,9 @@ test_progs() {
 		echo "test_progs:$?" >> "${STATUS_FILE}"
 		foldable end test_progs
 	fi
+}
 
+test_progs_noalu() {
 	foldable start test_progs-no_alu32 "Testing test_progs-no_alu32"
 	./test_progs-no_alu32 ${DENYLIST:+-d$DENYLIST} ${ALLOWLIST:+-a$ALLOWLIST} && true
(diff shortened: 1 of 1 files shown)


chantra:
LGTM

travis-ci/vmtest/run_selftests.sh
```diff
@@ -26,25 +26,31 @@ test_progs() {
 		echo "test_progs:$?" >> "${STATUS_FILE}"
 		foldable end test_progs
 	fi
```
mykola-lysenko, line 28: I assume it is correct, but will double check. Does it mean that we execute only no_alu32 tests each time? All other tests are being executed depending on the kernel version?

mykola-lysenko: E.g. on latest we execute all the tests, on 5.* we execute prog_tests and no_alu_32, and on 4.9.0 we execute only no_alu32

danielocfb: Yep, that seems to be the case.

danielocfb: That behavior is unchanged from earlier. E.g., check the selftest runs on https://github.com/libbpf/libbpf/pull/546.

mykola-lysenko: oh, I know, it is just something that caught my eye

mykola-lysenko (approved):

PR #549, opened 2022-08-17
Rename travis-ci/ directory to ci/

danielocfb:
We are no longer using Travis. As such, we should move away from a lot
of CI functionality located in a folder called travis-ci/. This change
renames the travis-ci/ directory to the more generic ci/.
To preserve backwards compatibility until all "consumers" have
transitioned, we add a symbolic link called travis-ci back. It will be
removed in the near term future.


travis-ci
```diff
@@ -0,0 +1 @@
+ci
```
mykola-lysenko, line 1: is this a symlink?

mykola-lysenko:
Bye bye travis ci!

travis-ci
mykola-lysenko: ok, missed your mention of it in the end, thank you!

mykola-lysenko (approved):

travis-ci
mykola-lysenko: I guess symlink will be required until travis-ci is removed in https://github.com/libbpf/ci/search?q=travis-ci

danielocfb: > I guess symlink will be required until travis-ci is removed in https://github.com/libbpf/ci/search?q=travis-ci

Yep, that is correct.

PR #553, opened 2022-08-18
Support running of individual tests

danielocfb:
This change adjusts the run_selftests.sh script to accept an optional
list of arguments specifying the tests to run. We will make use of it
once we run selftests in parallel.

### DIFF
--- a/ci/vmtest/run_selftests.sh
+++ b/ci/vmtest/run_selftests.sh
@@ -64,7 +64,13 @@ echo "ALLOWLIST: ${ALLOWLIST}"
 
 cd ${PROJECT_NAME}/selftests/bpf
 
-test_progs
-test_progs_noalu
-test_maps
-test_verifier
+if [ $# -eq 0 ]; then
+	test_progs
+	test_progs_noalu
+	test_maps
+	test_verifier
+else
+	for test_name in "$@"; do
+		"${test_name}"
+	done
+fi


mykola-lysenko:
not related to this change, but a general comment. Will this capability allow us to unite run_selftests.sh between vmtest and libbpf? I know the only difference is that we do not run all the tests for particular kernel versions and with this change we should be able to specify which tests to run for a version in the matrix and not in the shell script itself. Hope it explains my thought.

Well, we kinda can do this already, but with this change we can remove version checks from run_selftests.sh

ci/vmtest/run_selftests.sh
```diff
@@ -64,7 +64,13 @@ echo "ALLOWLIST: ${ALLOWLIST}"
...
-test_maps
-test_verifier
+if [ $# -eq 0 ]; then
+	test_progs
+	test_progs_noalu
+	test_maps
+	test_verifier
+else
+	for test_name in "$@"; do
+		"${test_name}"
```
mykola-lysenko, line 74: did not this will work, apparently it does

danielocfb:
> not related to this change, but a general comment. Will this capability allow us to unite run_selftests.sh between vmtest and libbpf? I know the only difference is that we do not run all the tests for particular kernel versions and with this change we should be able to specify which tests to run for a version in the matrix and not in the shell script itself. Hope it explains my thought.
> 
> Well, we kinda can do this already, but with this change we can remove version checks from run_selftests.sh

Yes, I had some vague thoughts in this realm and believe we could go down that route (and as you correctly point out, this change would be one step). I don't have immediate plans for unification of the two, but it's a thing we certainly should do eventually.

ci/vmtest/run_selftests.sh
danielocfb: Yep, it does work.

```sh
#!/bin/sh

test_verifier() {
  echo "running test_verifier"
}

echo "Trying to run: ${1}"
"${1}"
```

```
sh /tmp/test.sh test_verifier
> Trying to run: test_verifier
> running test_verifier
```

mykola-lysenko (approved):

PR #557, opened 2022-08-22
readme: add inverted logo version for dark background

qmonnet:
The original version of the logo and the accompanying text are not really readable on a dark background. This PR is a proposal to use an inverted version of the logo in that case.

Preview available on [the README in my branch](https://github.com/qmonnet/libbpf/tree/pr/logo-inverted).

imageimage

The proposed alternate logo is white and transparent. We lose the yellow part on the wings. This is just a suggestion, given that I suppose you have your own preferences on the matter, so let me know (or feel free to edit the PR). Instead of an inverted logo, we could also use a white border, or a coloured (yellow?) background.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,7 @@
-<img src="https://user-images.githubusercontent.com/508075/185997470-2f427d3d-f040-4eef-afc5-ae4f766615b2.png" width="40%" >
+<picture>
+  <source media="(prefers-color-scheme: light)" srcset="https://user-images.githubusercontent.com/508075/185997470-2f427d3d-f040-4eef-afc5-ae4f766615b2.png" width="40%">
+  <img src="https://user-images.githubusercontent.com/17001771/186033544-26fae355-1daa-42fb-bb46-46bbaa01caab.png" width="40%">
+</picture>
 
 libbpf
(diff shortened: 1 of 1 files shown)


anakryiko:
yeah, I knew it will be the problem but didn't yet have time to address this. Let me see if we can get a slightly better logo version to put on dark background.

anakryiko:
Ok, see https://github.com/libbpf/libbpf/pull/563. How do you like these variants:
Screen Shot 2022-08-24 at 10 11 18 PM
Screen Shot 2022-08-24 at 10 11 06 PM

qmonnet:
Looks great! Closer to the original, and the yellow border gives it a nice neon vibe.

I've also had something in store for bpftool for some time, trying to involve our graphic designer on it because my drawing skills are limited, but he's been busy. It will land eventually :)

anakryiko:
I've contracted a guy on fiverr.com to design the logo. My drawing skills are limited to simple geometrical shapes :) BTW, he also produced exactly the same all-white variant, which I put under https://github.com/libbpf/libbpf/tree/master/assets as well, just in case we'd need some mono-color print in the future or something along those lines.

PR #560, opened 2022-08-23
Use foldable helpers where applicable

danielocfb:
As discussed at some earlier point in time, some of the actions/workflow
logic does not use our foldable helpers despite being able to. Switch
them over.


mykola-lysenko (approved):

danielocfb:
Ah okay, so I missed that no coverity run was triggered by the change in question. These are failing now: https://github.com/libbpf/libbpf/runs/8022353803?check_suite_focus=true#step:4:18

PR #567, opened 2022-08-26
[s390x] support for consistent runner name + action runner update

chantra:
* use hostname to set the runner name
* updated https://github.com/actions/runner from c2.285.0 to v2.296.0


chantra:
I tested those changes on 2 runners, rebuilt the docker images and restarted the jobs:
image

danielocfb (approved):
Thanks for figuring this out! Looks good to me.

PR #570, opened 2022-08-29
Explicitly provide kernel-root argument to `prepare-rootfs` action

danielocfb:
Let's make the "kernel-root" explicit when using the `prepare-rootfs`
action, instead of relying on the default, `.kernel`.


mykola-lysenko (approved):
Change looks good, however CI is failing in the unrelated place for aarch64.

Please, retry after current run finishes

danielocfb:
> Change looks good, however CI is failing in the unrelated place for aarch64.
> 
> Please, retry after current run finishes

Thanks for the review. I did and it seems to have succeeded now.

PR #574, opened 2022-08-31
provide __PT_PARM6_REG.

liyan-ah:
Add r9d as __PT_PARM6_REG.

### DIFF
--- a/src/bpf_tracing.h
+++ b/src/bpf_tracing.h
@@ -111,6 +111,7 @@
 #define __PT_PARM3_REG rdx
 #define __PT_PARM4_REG rcx
 #define __PT_PARM5_REG r8
+#define __PT_PARM6_REG r9
 #define __PT_RET_REG rsp
 #define __PT_FP_REG rbp
 #define __PT_RC_REG rax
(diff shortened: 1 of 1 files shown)


liyan-ah:
I'm working on x86_64 architectures, and need to print 11 function parameters. found that arg6 (count from arg1) are not support by default, and only arg7 after can be got from stack. 
is there anything missed? or any other method to get arg6?
based on my work, I create this pr.

liyan-ah:
also, do we has any recommend method / function to get stack parameters?
found no documents

chenhengqi:
https://github.com/libbpf/libbpf#libbpf-and-general-bpf-usage-questions

liyan-ah:
@chenhengqi 
this confused me. only arg1 to arg5 are supported in the api document.

anakryiko:
@liyan-ah please see https://github.com/libbpf/libbpf-bootstrap/issues/111#issuecomment-1251917107, some architectures support more than 5 arguments in registers, so libbpf should provide macros (and BPF_KPROBE support) for those. Please consider submitting a patch for this.

Issue #577, opened 2022-09-02
Other dynamic libraries cannot link to libbpf.a


ChenYahui2019:
other dynamic library, such as `vpp af_xdp_plugin.so`, link to libbpf.a will meet errors below:
```
bin/ld: /usr/lib64/libbpf.a(libbpf.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
bin/ld: /usr/lib64/libbpf.a(netlink.o): relocation R_X86_64_32 against `.rodata.str1.8' can not be used when making a shared object; recompile with -fPIC
bin/ld: /usr/lib64/libbpf.a(nlattr.o): relocation R_X86_64_32 against `.rodata.str1.8' can not be used when making a shared object; recompile with -fPIC
bin/ld: /usr/lib64/libbpf.a(libbpf_probes.o): relocation R_X86_64_32S against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
```

I modify Makefile of libbpf, everything will be ok.
```
# git diff
diff --git a/src/Makefile b/src/Makefile
index 913082d..beb71ae 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -19,6 +19,7 @@ INCLUDES := -I. -I$(TOPDIR)/include -I$(TOPDIR)/include/uapi
 ALL_CFLAGS := $(INCLUDES)

 SHARED_CFLAGS += -fPIC -fvisibility=hidden -DSHARED
+STATIC_CFLAGS += -fPIC -shared

 CFLAGS ?= -g -O2 -Werror -Wall
 ALL_CFLAGS += $(CFLAGS) -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64
@@ -111,12 +112,11 @@ $(STATIC_OBJDIR) $(SHARED_OBJDIR):

 $(STATIC_OBJDIR)/%.o: %.c | $(STATIC_OBJDIR)
        $(call msg,CC,$@)
-       $(Q)$(CC) $(ALL_CFLAGS) $(CPPFLAGS) -c $< -o $@
+       $(Q)$(CC) $(ALL_CFLAGS) $(STATIC_CFLAGS) $(CPPFLAGS) -c $< -o $@

 $(SHARED_OBJDIR)/%.o: %.c | $(SHARED_OBJDIR)
        $(call msg,CC,$@)
[...]

PR #578, opened 2022-09-02
Other dynamic library link to libbpf.a will meet errors(https://githu…

ChenYahui2019:
Other dynamic library link to libbpf.a will meet errors(https://github.com/libbpf/libbpf/issues/577).
`-fPIC -shared` of CFLAG should be added when static library compiling.

Signed-off-by: Chen Yahui goodluckwillcomesoon@gmail.com


anakryiko:
This is not an appropriate change for normal use of static library. If you need to build static library with PIC, just provide EXTRA_CFLAGS with necessary extra flags.

PR #588, opened 2022-09-27
Adding network diag utils into actions-runner-libbpf container

CSXfb:
iproute2 and iputils-ping inside the container can be used to implement healthchecks or for troubleshooting.


chantra (approved):
Thanks!

PR #599, opened 2022-10-11
Add deny_namespace to DENYLIST-latest.s390x on CI

Decave:
The deny_namespace selftest was recently added to validate the userns_create LSM hooks. The test does not work on s390x because LSM uses BPF trampoline.
8206e4e95230 ("selftests/bpf: Add selftest deny_namespace to s390x deny list") was added to the bpf-next tree, but the bpf CI runs are still failing because the commit is not in that tree. So as to avoid merge conflicts, let's just blacklist it on libbpf CI.

Signed-off-by: David Vernet <void@manifault.com>

### DIFF
--- a/ci/vmtest/configs/DENYLIST-latest.s390x
+++ b/ci/vmtest/configs/DENYLIST-latest.s390x
@@ -1,4 +1,5 @@
 # TEMPORARY until bpf-next -> bpf merge
+deny_namespace                           # failed to attach: ERROR: strerror_r(-524)=22                                (trampoline)
 lru_bug                                  # prog 'printk': failed to auto-attach: -524
 
 # TEMPORARY


danielocfb:
Just wondering whether we actually need this setting here? It seems as if the CI system [uses](https://github.com/libbpf/libbpf/actions/runs/3221490763/jobs/5269489737#step:4:1659) bpf-next and I don't see any past failures due to this test. We surely should require it over in `vmtest` (which is used for `bpf` tests, e.g., https://github.com/kernel-patches/bpf/actions/runs/3220055049/jobs/5266514843#step:6:4949), though.

If I missed something, and we should merge it, I think we should remove the `selftests/bpf` prefix. It doesn't really have a meaning in this repository. Please also rephrase `DENYLIST.s390x` to `DENYLIST-latest.s390x` to more accurately reflect the file name in question. Lastly, and I know you landed a similar change upstream already, but in the future we may want to consider only deny listing the very sub-test failing (`deny_namespace/unpriv_userns_create_no_bpf`), and not the entire "suite" (I don't know if it makes a difference here and now, but it may, if folks add more sub-tests in the future).

Decave:
> Just wondering whether we actually need this setting here? It seems as if the CI system [uses](https://github.com/libbpf/libbpf/actions/runs/3221490763/jobs/5269489737#step:4:1659) bpf-next and I don't see any past failures due to this test. We surely should require it over in `vmtest` (which is used for `bpf` tests, e.g., https://github.com/kernel-patches/bpf/actions/runs/3220055049/jobs/5266514843#step:6:4949), though.

I'm not sure -- it may not be necessary. https://github.com/libbpf/libbpf/blob/master/ci/vmtest/configs/DENYLIST-latest.s390x#L1 does seem to indicate that tests are run for bpf as well? I am by no means an expert, so if this isn't necessary we can just drop this.

> If I missed something and we should merge it, I think we should remove the `selftests/bpf` prefix. It doesn't really have a meaning in this repository. Please also rephrase `DENYLIST.s390x` to `DENYLIST-latest.s390x` to more accurately reflect the file name in question.

Will do.

>Lastly, and I know you landed a similar change upstream already, but in the future we may want to consider only deny listing the very sub-test failing (`deny_namespace/unpriv_userns_create_no_bpf`), and not the entire "suite" (I don't know if it makes a difference here and now, but it may, if folks add more sub-tests in the future).

I tend to agree, though I think that discussion is quite the outside the scope here and should probably be discussed on the bpf@vger list and upstreamed if we want to support it. Or, is specifying sub-tests already supported? I do agree that it'd be useful, speaking for myself, I've written test suites that only need to black list one or two testcases.

Decave:
Addressed @danielocfb's feedback. If this PR ends up not being necessary, we can just drop it.

anakryiko:
libbpf CI is testing against bpf-next only, so yeah, we don't need this in libbpf/libbpf.

danielocfb:
> > Just wondering whether we actually need this setting here? It seems as if the CI system [uses](https://github.com/libbpf/libbpf/actions/runs/3221490763/jobs/5269489737#step:4:1659) bpf-next and I don't see any past failures due to this test. We surely should require it over in `vmtest` (which is used for `bpf` tests, e.g., https://github.com/kernel-patches/bpf/actions/runs/3220055049/jobs/5266514843#step:6:4949), though.
> 
> I'm not sure -- it may not be necessary. https://github.com/libbpf/libbpf/blob/master/ci/vmtest/configs/DENYLIST-latest.s390x#L1 does seem to indicate that tests are run for bpf as well?

Indeed, that seems confusing. Will remove it.

> > Lastly, and I know you landed a similar change upstream already, but in the future we may want to consider only deny listing the very sub-test failing (`deny_namespace/unpriv_userns_create_no_bpf`), and not the entire "suite" (I don't know if it makes a difference here and now, but it may, if folks add more sub-tests in the future).
> 
> I tend to agree, though I think that discussion is quite the outside the scope here and should probably be discussed on the bpf@vger list and upstreamed if we want to support it. Or, is specifying sub-tests already supported? I do agree that it'd be useful, speaking for myself, I've written test suites that only need to black list one or two testcases.

Yep, it should already be supported.

PR #600, opened 2022-10-11
Remove lru_bug from DENYLIST-latest.s390x

danielocfb:
The comment associated with the entry is a bit confusing. It stemmed from the test being denylisted on `bpf`, but not `bpf-next` in the past. Regardless, by now said change has propagated to both trees, so we no longer need to carry around this deny list entry here.


chantra (approved):
Errors seem to be unrelated.

PR #612, opened 2022-11-12
Libbpf sync 2022-11-12T21:07:43.458Z

anakryiko:


anakryiko:
@teknoraver https://github.com/libbpf/libbpf/commit/1c17672353a3d6c88f4c0e260ba45a86dc803fed actually broke the sync script. I had to debug that and undo those changes (and unfortunately I forgot that we just recently added those quotes, so it took more time). Anyways, I think shellcheck is wrong (and https://github.com/koalaman/shellcheck/wiki/SC2068 incorrectly states that when double-quoting arrays each element is expanded, that's not what's happening in practice, entire array contents actually becomes a single-quoted string). So the question I have is if there is any way to disable SC2068 check specifically?

If not, I'll have to disable shellcheck step, unfortunately

chantra:
> Anyways, I think shellcheck is wrong (and
> https://github.com/koalaman/shellcheck/wiki/SC2068 incorrectly states
> that when double-quoting arrays each element is expanded, that's not what's
> happening in practice, entire array contents actually becomes a
> single-quoted string).
>

That should not be the case. “$@“ should expand as individual arguments.

So the question I have is if there is any way to disable SC2068 check
> specifically?
>
> If not, I'll have to disable shellcheck step, unfortunately
>

You can add a stanza at the top of the file as explained in
https://www.shellcheck.net/wiki/Ignore

anakryiko:
> Anyways, I think shellcheck is wrong (and https://github.com/koalaman/shellcheck/wiki/SC2068 incorrectly states that when double-quoting arrays each element is expanded, that's not what's happening in practice, entire array contents actually becomes a single-quoted string).
> That should not be the case. “$@“ should expand as individual arguments.

Maybe, but that's what I see in practice (by setting `set -x`):

```
$ git rev-list --no-merges --topo-order --reverse libbpf-baseline-2022-11-12T20-58-52.731Z..libbpf-tip-2022-11-12T20-58-52.731Z 'tools/include/uapi/linux/if_link.h tools/include/uapi/linux/pkt_sched.h tools/include/uapi/linux/netlink.h tools/lib/bpf tools/include/uapi/linux/if_xdp.h tools/include/uapi/linux/pkt_cls.h tools/include/uapi/linux/btf.h Documentation/bpf/libbpf tools/include/uapi/linux/bpf.h include/uapi/linux/perf_event.h tools/include/uapi/linux/bpf_common.h :^tools/lib/bpf/Makefile :^tools/lib/bpf/Build :^tools/lib/bpf/.gitignore :^tools/include/tools/libc_compat.h'
```

That list of files shouldn't be a single string argument, but it is.

> So the question I have is if there is any way to disable SC2068 check
> specifically? If not, I'll have to disable shellcheck step, unfortunately
> You can add a stanza at the top of the file as explained in https://www.shellcheck.net/wiki/Ignore

Cool, thanks!

chantra:
> Maybe, but that's what I see in practice (by setting set -x):
>
> $ git rev-list --no-merges --topo-order --reverse libbpf-baseline-2022-11-12T20-58-52.731Z..libbpf-tip-2022-11-12T20-58-52.731Z 'tools/include/uapi/linux/if_link.h tools/include/uapi/linux/pkt_sched.h tools/include/uapi/linux/netlink.h tools/lib/bpf tools/include/uapi/linux/if_xdp.h tools/include/uapi/linux/pkt_cls.h tools/include/uapi/linux/btf.h Documentation/bpf/libbpf tools/include/uapi/linux/bpf.h include/uapi/linux/perf_event.h tools/include/uapi/linux/bpf_common.h :^tools/lib/bpf/Makefile :^tools/lib/bpf/Build :^tools/lib/bpf/.gitignore :^tools/include/tools/libc_compat.h'
>
>
> That list of files shouldn't be a single string argument, but it is.
>

Ok. The problem seems to be that LIBBPF_PATHS is actually a string, not an
array:
https://github.com/libbpf/libbpf/blob/971ad8f8d0c92434a0ef6693c75c38326ea07258/scripts/sync-kernel.sh#L54
So indeed, it relies on expanding the “array” to have each of those element
in the string become an individual element.

I think something like:
LIBBPF_PATHS=("${!PATH_MAP[@]}" ":^tools/lib/bpf/Makefile"
":^tools/lib/bpf/Build" ":^tools/lib/bpf/.gitignore"
":^tools/include/tools/libc_compat.h")

teknoraver:
> So the question I have is if there is any way to disable SC2068 check specifically?
> 
> If not, I'll have to disable shellcheck step, unfortunately

Sure, we can disable every warning separately. In `.github/workflows/lint.yml` just add `--exclude=SC2068`

teknoraver:
@anakryiko @chantra 

> Anyways, I think shellcheck is wrong (and https://github.com/koalaman/shellcheck/wiki/SC2068 incorrectly states that when double-quoting arrays each element is expanded, that's not what's happening in practice, entire array contents actually becomes a single-quoted string).
> That should not be the case. “$@“ should expand as individual arguments. So the question I have is if there is any way to disable SC2068 check
> specifically? If not, I'll have to disable shellcheck step, unfortunately
> You can add a stanza at the top of the file as explained in https://www.shellcheck.net/wiki/Ignore

Well, that's what really happens, “$@“ is expanded in single arguments.
This snippet:
```bash
array=(a b "c d" ef)

echo "Without quotes:"
for i in ${array[@]}; do
	echo "'$i'"
done

echo "With quotes:"
for i in "${array[@]}"; do
	echo "'$i'"
done
```
Will run as:
```
$ bash array.sh
Without quotes:
'a'
'b'
'c'
'd'
'ef'
With quotes:
'a'
'b'
'c d'
'ef'
```

In our case, the problem (I think) is that in https://github.com/libbpf/libbpf/blob/v1.0.1/scripts/sync-kernel.sh#L54 `LIBBPF_PATHS` is declared as a unique string instead of an array:

`LIBBPF_PATHS="${!PATH_MAP[@]} :^tools/lib/bpf/Makefile :^tools/lib/bpf/Build :^tools/lib/bpf/.gitignore :^tools/include/tools/libc_compat.h"`

So it must be either treated as a simple string with `$LIBBPF_PATHS` or converted to an array in the declaration.

anakryiko:
@chantra, @teknoraver thanks for investigating, I'll update LIBBPF_PATHS and LIBBBPF_VIEW_PATHS to be "real" array then. It's amazing that [@] thing works on strings, though.

PR #614, opened 2022-11-14
sync: make LIBBPF_PATHS and LIBBPF_VIEW_PATHS into real array variables

anakryiko:
Use correct Bash syntax to define these two variables as arrays. Drop shellcheck opt-out for unquoted use of array.


anakryiko:
cc @teknoraver, @chantra

teknoraver (approved):

PR #620, opened 2022-11-16
ci: fix Debian builds due to pkg-config dependency change

anakryiko:
Seems like we need pkgconfig dependency instead of pkg-config.


ci/managers/debian.sh
```diff
@@ -6,7 +6,7 @@ CONT_NAME="${CONT_NAME:-libbpf-debian-$DEBIAN_RELEASE}"
 ENV_VARS="${ENV_VARS:-}"
 DOCKER_RUN="${DOCKER_RUN:-docker run}"
 REPO_ROOT="${REPO_ROOT:-$PWD}"
-ADDITIONAL_DEPS=(clang pkg-config gcc-10)
+ADDITIONAL_DEPS=(clang pkgconfig gcc-10)
```
chantra, line 9: [`pkgconf`](https://packages.debian.org/testing/devel/pkgconf) :)

PR #494, opened 2022-05-09
build: Enable -fPIC for static library builds

mestery:
This commit moves -fPIC from the SHARED_CFLAGS options into the base
CFLAGS, allowing the resulting static libbpf library to be linked into
other shared libraries.

Closes #493

Signed-off-by: Kyle Mestery <mestery@mestery.com>

PR #631, opened 2022-11-24
Add -fPIC to CFLAGS

tjcw:
I found that xdp-project/xdp-tools would not build in support of xdp-project/bpf-examples. The reason for this turned out to be that the commit level of libxdp in use by xdp-project/xdp-tools was trying to build a shared libxdp on top of an unshared libbpf.
For this to work it is necessary for the code in the unshared libbpf to be built with the '-fPIC' flag.
Adding '-fPIC' has no functional disadvantage; the only downside is that the code might be infinitesimally slower.


anakryiko:
please see https://github.com/libbpf/libbpf/pull/102 and https://github.com/libbpf/libbpf/pull/494. Why doesn't xdp-tools build libbpf.a with -fPIC ?

tjcw:
The specific problem was that libxdp was being built as a shared library and libbpf was being built as an archive library
 So the libxdp was getting a copy of the libbpf code and trying to build it into the shared code.

anakryiko:
> The specific problem was that libxdp was being built as a shared library and libbpf was being built as an archive library So the libxdp was getting a copy of the libbpf code and trying to build it into the shared code.

I'm not sure I understood the answer, sorry. Is it possible to change libxdp's Makefile to use shared library version of libbpf if libxdp is built as shared library? Because that seems like a sensible way to go about this.

tjcw:
Yes, that should be a sensible way forward. Who is maintainer for libxdp ?

anakryiko:
cc @tohojo

tohojo:
libxdp deals just fine with linking against libbpf as either a shared or static library. And the build system in the bpf-examples repository deliberately configures both to be linked statically anyway. So yeah, agree, there's no bug in libbpf here...

PR #636, opened 2022-12-06
ci: list tests to run explicitly

anakryiko:
Only allow test_progs, test_progs-no_alu32, and test_verifier. We don't like test_maps.

### DIFF
--- a/.github/actions/vmtest/action.yml
+++ b/.github/actions/vmtest/action.yml
@@ -94,6 +94,7 @@ runs:
       uses: libbpf/ci/run-qemu@master
       with:
         arch: ${{ inputs.arch }}
+        kernel-test: 'test_progs,test_progs_no_alu32,test_verifier'
         img: '/tmp/root.img'
         vmlinuz: 'vmlinuz'
         kernel-root: '.kernel'


danielocfb (approved):
Fuzzing tests seem to be broken :-(
Looks good assuming all tests pass. Would be valuable to explain more of the rationale than just saying "we don't like ..." in the description, though.

evverx:
> Fuzzing tests seem to be broken :-(

CIFuzz was broken at the time but that regression was fixed in https://github.com/google/oss-fuzz/issues/9151.

anakryiko:
@chantra I tried to use explicit list of tests, but that doesn't seem to work on libbpf CI. Is there something more that needs to be done?

For now I landed a separate commit which just commented out `./test_maps` run, but if you can check what I did wrong here I'd appreciate it!

chantra:
> @chantra I tried to use explicit list of tests, but that doesn't seem to work on libbpf CI. Is there something more that needs to be done?
> 
> For now I landed a separate commit which just commented out `./test_maps` run, but if you can check what I did wrong here I'd appreciate it!

vmtest are sourced from the [projects' repo themselves](https://github.com/libbpf/ci/blob/524c4d992d54927ee50327cc085416f60fb1efbc/prepare-rootfs/run_vmtest.sh#L25). You would need to port the logic in https://github.com/kernel-patches/vmtest/commit/e5bfcc764374556911e8de7fe22b547fe4983202 to https://github.com/libbpf/libbpf/blob/master/ci/vmtest/run_selftests.sh

Making this more transparent than copying the logic over does not seem that easy/straightforward.

PR #657, opened 2023-02-17
README: Clarification on how to handle header files in include/

liu-song-6:
The header files should not be reduced versions. And they should be sync'ed by sync-kernel.sh.

Also, clarify that this rule applies to headers in include/uapi/linux/.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,12 +111,12 @@ kernel. You'll need:
   - kernel built with `CONFIG_DEBUG_INFO_BTF=y` option;
   - you can check if your kernel has BTF built-in by looking for
     `/sys/kernel/btf/vmlinux` file:
-  
+
 ```shell
 $ ls -la /sys/kernel/btf/vmlinux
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -175,10 +175,11 @@ bpf-next to Github sync
 All the gory details of syncing can be found in `scripts/sync-kernel.sh`
 script.
 
-Some header files in this repo (`include/linux/*.h`) are reduced versions of
```
anakryiko, line 178: hold on, here we are talking about kernel-internal include/linux/*.h headers, not UAPI headers

anakryiko (changes requested):
README talks about kernel-internal files. We can add a separate point about UAPI headers copied as-is, but it should be in addition to everything there is.

README.md
liu-song-6: Aha, I guess I misunderstood this paragraph. Maybe another paragraph about include/uapi/linux/*.h?

liu-song-6: Sigh, I keep missing comments on GitHub. :(

anakryiko:
closing for now

PR #663, opened 2023-03-06
sync: Add section about need for Makefile adjustments

danielocfb:
When performing a sync with the kernel repository using the `sync-kernel.sh` script, it may be necessary to manually adjust the library's `Makefile` if:
- new source files were added upstream
- new public headers were added upstream

This change adjusts `SYNC.md` to spell out this need.

### DIFF
--- a/SYNC.md
+++ b/SYNC.md
@@ -152,6 +152,20 @@ sure that each such commit has `Signed-off-by: Your Full Name <your@email.com>`,
 just like you'd do that for Linux upstream patch. Libbpf closely follows kernel
 conventions and styling, so please help maintaining that.
 
+Including new sources
+---------------------
+
+If entirely new source files (typically `*.c`) were added to the library in the
+kernel repository, it may be necessary to add these to the build system
+manually (you may notice linker errors otherwise), because the script cannot
+handle such changes automatically. To that end, edit `src/Makefile` as
+necessary. Commit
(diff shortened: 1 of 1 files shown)


SYNC.md
```diff
@@ -112,6 +112,12 @@ If sync is successful, your `~/linux` repo will be left in original state on
 the original HEAD commit. `~/libbpf` repo will now be on a new branch, named
 `libbpf-sync-<timestamp>` (e.g., `libbpf-sync-2023-02-28T00-53-40.072Z`).
 
+If entirely new source files were added to the library in the kernel repository,
```
anakryiko, line 115: Wording looks good, but I'm afraid people will mostly skim over this anyways. So let's add this as "common issues" subsection, like we did with "Updating allow/deny lists" and "Regenerating vmlinux.h header", as this would most certainly be helpful *after* CI build break.

Let's also use a link to your commit (https://github.com/libbpf/libbpf/commit/c2495832ced4239bcd376b9954db38a6addd89ca) as an example, ok?


And separate from all of this, we need to have a better test to catch missing public header APIs and also missing.o (current build test won't detect this if added .o file is not used from any other libbpf .c file, we got lucky with zip.c, it could have went unnoticed if it was containing only public-facing functionality). Let's chat about this separately, maybe you can help figuring this out.

danielocfb: Sure, done.

anakryiko:
Looks great, thanks!

PR #680, opened 2023-04-25
libbtf: allow 128B sized BTF_KIND_INT for LLVM DW_ATE_unsigned_1024

nickdesaulniers:
Similar to pahole
commit 7d8e829f636f ("btf_encoder: Sanitize non-regular int base type")

We can see some unusual sizes in DW_AT_location expressions, example:
```
0x0e4559e2:     DW_TAG_variable
                  DW_AT_location	(indexed (0xcf) loclist = 0x00436e0e:
                     [0x000000000000002c, 0x0000000000000038): DW_OP_breg8 W8+0, DW_OP_convert (0x0e4360cb) "DW_ATE_unsigned_1024", DW_OP_convert (0x0e4360c6) "DW_ATE_unsigned_32", DW_OP_lit3, DW_OP_and, DW_OP_stack_value, DW_OP_piece 0x8)
                  DW_AT_name	("reg_value")
                  DW_AT_type	(0x0e436189 "uint32_t")
```

The following LLVM IR can reproduce the issue:
```llvm
define i8 @foo() !dbg !7 {
entry:
  call void @llvm.dbg.value(metadata i8 32, metadata !9, metadata !DIExpression(DW_OP_plus, DW_OP_LLVM_convert, 8, DW_ATE_signed, DW_OP_LLVM_convert, 1024, DW_ATE_unsigned, DW_OP_stack_value)), !dbg !11
  ret i8 0, !dbg !12
}

declare void @llvm.dbg.value(metadata, metadata, metadata)

!llvm.dbg.cu = !{!0}
!llvm.module.flags = !{!5, !6}

!0 = distinct !DICompileUnit(language: DW_LANG_C99, file: !1, producer: "clang version 9.0.0 ", isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug, enums: !2, nameTableKind: None)
!1 = !DIFile(filename: "dbg-foo.c", directory: "/tmp", checksumkind: CSK_MD5, checksum: "b35f80a032deb2a30bc187d564b5a775")
!2 = !{}
!4 = !DIFile(filename: "dbg-bar.c", directory: "/tmp", checksumkind: CSK_MD5, checksum: "9836bb594260d883960455e7d8bc51ea")
!5 = !{i32 2, !"Dwarf Version", i32 5}
!6 = !{i32 2, !"Debug Info Version", i32 3}
!7 = distinct !DISubprogram(name: "foo", scope: !1, file: !1, line: 7, type: !8, scopeLine: 8, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !0, retainedNodes: !2)
!8 = !DISubroutineType(types: !2)
!9 = !DILocalVariable(name: "y", scope: !7, file: !1, line: 9, type: !10)
!10 = !DIBasicType(name: "int", size: 32, encoding: DW_ATE_signed)
!11 = !DILocation(line: 9, column: 14, scope: !7)
!12 = !DILocation(line: 10, column: 3, scope: !7)
```
```sh
$ llc -filetype=obj -O0 < reduced.ll
$ pahole -J reduced.o
[2] INT DW_ATE_unsigned_1024 Error emitting BTF type
Encountered error while encoding BTF.
```

Link: https://git.kernel.org/pub/scm/devel/pahole/pahole.git/commit/?id=7d8e829f636f47aba2e1b6eda57e74d8e31f733c

### DIFF
--- a/src/btf.c
+++ b/src/btf.c
@@ -1763,7 +1763,7 @@ int btf__add_int(struct btf *btf, const char *name, size_t byte_sz, int encoding
 	if (!name || !name[0])
 		return libbpf_err(-EINVAL);
 	/* byte_sz must be power of 2 */
-	if (!byte_sz || (byte_sz & (byte_sz - 1)) || byte_sz > 16)
+	if (!byte_sz || (byte_sz & (byte_sz - 1)) || byte_sz > 128)
 		return libbpf_err(-EINVAL);
 	if (encoding & ~(BTF_INT_SIGNED | BTF_INT_CHAR | BTF_INT_BOOL))
 		return libbpf_err(-EINVAL);


nickdesaulniers:
This was reported by an Android vendor; I haven't yet been able to get my hands on a C based reproducer, but I was able to hack up then reduce LLVM IR to produce this same `DW_ATE_unsigned_1024` via `DW_OP_LLVM_convert`.

I wasn't sure how to run tests to see what else this affects, or if the results can be dumped (I'm not familiar with most btf tooling).

nickdesaulniers:
cc @yonghong-song

eddyz87:
The BTF [doc](https://www.kernel.org/doc/html/latest/bpf/btf.html#btf-kind-int) says the following:

> 2.2.1 BTF_KIND_INT
> ...
> btf_type is followed by a u32 with the following bits arrangement:
> ```c
> #define BTF_INT_ENCODING(VAL)   (((VAL) & 0x0f000000) >> 24)
> #define BTF_INT_OFFSET(VAL)     (((VAL) & 0x00ff0000) >> 16)
> #define BTF_INT_BITS(VAL)       ((VAL)  & 0x000000ff)
> ```
> ...
> The BTF_INT_BITS() specifies the number of actual bits held by this int type. For example, a 4-bit bitfield encodes BTF_INT_BITS() equals to 4. The btf_type.size * 8 must be equal to or greater than BTF_INT_BITS() for the type. The maximum value of BTF_INT_BITS() is 128.

So, accommodating a 1024 bit integer would require some BTF format changes.

Do you hit this error from pahole or from some other tooling?
Do you have means to check if this `DW_ATE_unsigned_1024` is referenced only from `DW_AT_location` or there are some "real" usages of this type (e.g. as a `DW_AT_type`)?

yonghong-song:
Right, currently, the maximum integer bytes is 16 (for __int128). See tools/lib/bpf/btf.c:
```
int btf__add_int(struct btf *btf, const char *name, size_t byte_sz, int encoding)
{
        struct btf_type *t;
        int sz, name_off;

        /* non-empty name */
        if (!name || !name[0])
                return libbpf_err(-EINVAL);
        /* byte_sz must be power of 2 */
        if (!byte_sz || (byte_sz & (byte_sz - 1)) || byte_sz > 16)
                return libbpf_err(-EINVAL);

```

Let us add the similar condition to pahole as well like below patch,
```
diff --git a/btf_encoder.c b/btf_encoder.c
index 65f6e71..1aa0ad0 100644
--- a/btf_encoder.c
+++ b/btf_encoder.c
@@ -394,7 +394,7 @@ static int32_t btf_encoder__add_base_type(struct btf_encoder *encoder, const str
         * these non-regular int types to avoid libbpf/kernel complaints.
         */
        byte_sz = BITS_ROUNDUP_BYTES(bt->bit_size);
-       if (!byte_sz || (byte_sz & (byte_sz - 1))) {
+       if (!byte_sz || (byte_sz & (byte_sz - 1)) || byte_sz > 16) {
                name = "__SANITIZED_FAKE_INT__";
                byte_sz = 4;
        }
```

@nickdesaulniers @eddyz87 what do you think?

yonghong-song:
Also, it would be good to have a C reproducer to generate DW_ATE_unsigned_1024 so we can understand how this DW_ATE_unsigned_1024 is generated to ensure we don't need anything.

eddyz87:
> Right, currently, the maximum integer bytes is 16 (for __int128). See tools/lib/bpf/btf.c:
> 
> ```
> int btf__add_int(struct btf *btf, const char *name, size_t byte_sz, int encoding)
> {
>         struct btf_type *t;
>         int sz, name_off;
> 
>         /* non-empty name */
>         if (!name || !name[0])
>                 return libbpf_err(-EINVAL);
>         /* byte_sz must be power of 2 */
>         if (!byte_sz || (byte_sz & (byte_sz - 1)) || byte_sz > 16)
>                 return libbpf_err(-EINVAL);
> ```
> 
> Let us add the similar condition to pahole as well like below patch,
> 
> ```
> diff --git a/btf_encoder.c b/btf_encoder.c
> index 65f6e71..1aa0ad0 100644
> --- a/btf_encoder.c
> +++ b/btf_encoder.c
> @@ -394,7 +394,7 @@ static int32_t btf_encoder__add_base_type(struct btf_encoder *encoder, const str
>          * these non-regular int types to avoid libbpf/kernel complaints.
>          */
>         byte_sz = BITS_ROUNDUP_BYTES(bt->bit_size);
> -       if (!byte_sz || (byte_sz & (byte_sz - 1))) {
> +       if (!byte_sz || (byte_sz & (byte_sz - 1)) || byte_sz > 16) {
>                 name = "__SANITIZED_FAKE_INT__";
>                 byte_sz = 4;
>         }
> ```
> 
> @nickdesaulniers @eddyz87 what do you think?

Yes, either this or teach pahole to ignore types that are referenced only from "auxiliary" places like `DW_AT_location`.
(should probably be discussed in the dwarves mailing list).

nickdesaulniers:
> @nickdesaulniers @eddyz87 what do you think?

Sure that WFM.

> Yes, either this or teach pahole to ignore types that are referenced only from "auxiliary" places like DW_AT_location.
(should probably be discussed in the dwarves mailing list).

Yeah, I'm curious why pahole is validating the loclists at all. Does pahole need to know about stack slots?

> Also, it would be good to have a C reproducer to generate DW_ATE_unsigned_1024 so we can understand how this DW_ATE_unsigned_1024 is generated to ensure we don't need anything.

I agree; I'm working with the vendor to show them how to run creduce, but I can't promise anything.

yonghong-song:
btf does not need or encode location info. But currently pahole process dwarf types in sequence which will allocate 'id' for these types. For example, below are dwarf output for the above example,

```
.debug_info contents:
0x00000000: Compile Unit: length = 0x0000004f, format = DWARF32, version = 0x0005, unit_type = DW_UT_compile, abbr_offset = 0x0000, addr_size = 0x08 (next unit at 0x00000053)

0x0000000c: DW_TAG_compile_unit
              DW_AT_producer    ("clang version 9.0.0 ")
              DW_AT_language    (DW_LANG_C99)
              DW_AT_name        ("dbg-foo.c")
              DW_AT_str_offsets_base    (0x00000008)
              DW_AT_stmt_list   (0x00000000)
              DW_AT_comp_dir    ("/tmp")
              DW_AT_low_pc      (0x0000000000000000)
              DW_AT_high_pc     (0x0000000000000003)
              DW_AT_addr_base   (0x00000008)

0x00000023:   DW_TAG_base_type
                DW_AT_name      ("DW_ATE_signed_8")
                DW_AT_encoding  (DW_ATE_signed)
                DW_AT_byte_size (0x01)

0x00000027:   DW_TAG_base_type
                DW_AT_name      ("DW_ATE_unsigned_1024")
                DW_AT_encoding  (DW_ATE_unsigned)
                DW_AT_byte_size (0x80)

0x0000002b:   DW_TAG_subprogram
                DW_AT_low_pc    (0x0000000000000000)
                DW_AT_high_pc   (0x0000000000000003)
                DW_AT_frame_base        (DW_OP_reg7 RSP)
                DW_AT_name      ("foo")
                DW_AT_decl_file ("/tmp/dbg-foo.c")
                DW_AT_decl_line (7)
                DW_AT_prototyped        (true)
                DW_AT_external  (true)

0x00000036:     DW_TAG_variable
                  DW_AT_location        (DW_OP_constu 0x20, DW_OP_plus, DW_OP_convert (0x00000023) "DW_ATE_signed_8", DW_OP_convert (0x00000027) "DW_ATE_unsigned_1024", DW_OP_stack_value)
                  DW_AT_name    ("y")
                  DW_AT_decl_file       ("/tmp/dbg-foo.c")
                  DW_AT_decl_line       (9)
                  DW_AT_type    (0x0000004e "int")

0x0000004d:     NULL

0x0000004e:   DW_TAG_base_type
                DW_AT_name      ("int")
                DW_AT_encoding  (DW_ATE_signed)
                DW_AT_byte_size (0x04)

0x00000052:   NULL
```

pahole will process the following type and allocate a id for it.
```
0x00000027:   DW_TAG_base_type
                DW_AT_name      ("DW_ATE_unsigned_1024")
                DW_AT_encoding  (DW_ATE_unsigned)
                DW_AT_byte_size (0x80)
```

To avoid messing up id mapping, btf has to encode it.

The only thing we can do is to filter out this type in dwarf_loader before allocating a id. 
We could do this by detecting whether eventual coding is btf or not. If eventually for btf encoding, we can ignore
such a type as btf encoding will fail any way. This could be an option, but currently dwarf_loader.c does not check
eventual encoding (btf, ctf, etc.) and the control is through config options.

That is why I want to get a source code so I can understand when such *large* int types are generated so I can guess how many such types could be in the dwarf.

Let me prepare a patch and send to pahole mailing list for a discussion.

nickdesaulniers:
https://lore.kernel.org/dwarves/20230426055030.3743074-1-yhs@fb.com/

dwblaikie:
> Do you have means to check if this DW_ATE_unsigned_1024 is referenced only from DW_AT_location or there are some "real" usages of this type (e.g. as a DW_AT_type)?

While it's maybe not the best thing to /rely/ on (it'd probably be good if these DWARF parsing tools walked the `DW_AT_type` references to determine which types they need to know, rather than parsing all of them unconditionally - so as to avoid future problems like this one), it is the case that these specific types named "DW_AT_unsigned_<size>" only come from LLVM's DWARF generation for location expressions here: https://github.com/llvm/llvm-project/blob/5202aa18f6fea8b4864906ce4846545c7980d723/llvm/lib/CodeGen/AsmPrinter/DwarfCompileUnit.cpp#L1619

PR #686, opened 2023-05-01
Fix comment about arc provides user_regs_struct

nak3:
This patch makes tiny changes by:
- Fixing the comment about `user_pt_regs` not `user_pt_regs`.
- Adding the comment for riscv about `user_pt_regs`.


anakryiko:
please see https://github.com/libbpf/libbpf#libbpf, all the patches for libbpf source should go through Linux bpf-next tree, which is authoritative source of libbpf; Github is mostly just a mirror.

nak3:
Ah! Thank you so much.
I will send the PR against bpf-next.

PR #692, opened 2023-05-24
git: make .gitattributes compatible with git-archive-all action

anakryiko:
As reported by Quentin, using Github Action to archive all submodules (e.g., for retsnoop release packaging) is impacted by it not supporting "<glob>/" pattern in .gitattributes. Use "<glob>/**" instead.

  [0] https://github.com/anakryiko/retsnoop/pull/42#issuecomment-1560797837

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-assets/ export-ignore
+assets/** export-ignore


qmonnet (approved):
Thanks! I tested, and confirm that this fix the ignore rules for `git_archive_all.py`.

For the context, the documentation for the syntax for `gitattributes` says that:

> patterns that match a directory do not recursively match paths inside that directory (so using the trailing-slash path/ syntax is pointless in an attributes file; use path/** instead)

We can check that the rule is not supposed to apply to the files under the directory used as a pattern:

```console
$ git check-attr -a libbpf/assets/libbpf-logo-compact-darkbg.png
<no result>
```

Based on that, `git_archive_all.py` does not apply the rule for pattern `assets/` (internally it runs `git check-attr` on each file from `git ls-files` to check whether it should be excluded, so it just doesn't get the attribute).

However, at some point `git archive` got the ability to work with this pattern (https://github.com/git/git/commit/5ff247ac0cc9b2d09c8c24bd0c8df12eed94aebc), even though this contradicts the docs and the behaviour from `git check-attr`.

By changing the pattern to `assets/**`, we observe that the attribute is set indeed:

```console
$ git check-attr -a libbpf/assets/libbpf-logo-compact-darkbg.png
libbpf/assets/libbpf-logo-compact-darkbg.png: export-ignore: set
```

And `git_archive_all.py` works as expected.

PR #701, opened 2023-06-25
libbpf: provide num present cpus

carofe82:
It allows tools to iterate over CPUs present
in the system that are actually running processes, which can be less than the number of possible CPUs.


carofe82:
The goal of this PR is to provide a way for bpf-cpufreq to get the number of present cpu to fix this error:
https://github.com/iovisor/bcc/issues/4651

My plan is that when this PR is approved, I'll create a PR for bpf-cpufreq. I can work on other approaches as well if preferred by the maintainers.

carofe82:
Closing PR after realizing it should have been done via the mailing list.

PR #712, opened 2023-07-31
bpf: use local includes inside the library

sakateka:
Today there is no way to include library headers without doing `PREFIX=<some tmp dir> make install` with some prefix to try a new version of the library without installing it globally, my fix allows to do this. And it seems to me that this does not violate the current behavior.


sakateka:
Thanks, of course I read this text, but the habit of using github blinded me.

PR #718, opened 2023-08-24
fuzz: use https-based URL for elfutils

anakryiko:
For environments behind proxies, having https:// URL for pulling GIT is more convenient.


anakryiko:
@evverx any objections?

evverx:
I think `https` is better. I'm not sure how `git` ended up there :-)

PR #723, opened 2023-08-29
Fix bpf_core_read.h comment wrt bpf_core_relo struct

### LINKED ISSUE #722: bpf_core_read comment needs an update
https://github.com/libbpf/libbpf/blob/5a46421ad837e876197295844696884c8587852a/src/bpf_core_read.h#L234C1-L239C73

```
 * __builtin_preserve_access_index() takes as an argument an expression of
 * taking an address of a field within struct/union. It makes compiler emit
 * a relocation, which records BTF type ID describing root struct/union and an
 * accessor string which describes exact embedded field that was used to take
 * an address. See detailed description of this relocation format and
 * semantics in comments to struct bpf_field_reloc in libbpf_internal.h.
```

However, there doesn't seem to be such a thing as `struct bpf_field_reloc`.

Some commits, like 3bde6ca8e85ac09a966cc891b0c9c97a338ce0aa seem to have removed the last vestiges of that struct... maybe it's called `bpf_core_relo` now?

dimaqq:
Fix #722

### DIFF
--- a/src/bpf_core_read.h
+++ b/src/bpf_core_read.h
@@ -236,7 +236,7 @@ enum bpf_enum_value_kind {
  * a relocation, which records BTF type ID describing root struct/union and an
  * accessor string which describes exact embedded field that was used to take
  * an address. See detailed description of this relocation format and
- * semantics in comments to struct bpf_field_reloc in libbpf_internal.h.
+ * semantics in comments to struct bpf_core_relo in include/uapi/linux/bpf.h.
  *
  * This relocation allows libbpf to adjust BPF instruction to use correct
  * actual field offset, based on target kernel BTF type that matches original


danielocfb:
As per:
> Libbpf authoritative source code is developed as part of [bpf-next Linux source tree](https://kernel.googlesource.com/pub/scm/linux/kernel/git/bpf/bpf-next) under tools/lib/bpf subdirectory and is periodically synced to Github. As such, all the libbpf changes should be sent to [BPF mailing list](http://vger.kernel.org/vger-lists.html#bpf), please don't open PRs here unless you are changing Github-specific parts of libbpf (e.g., Github-specific Makefile).

you may want to submit this patch to the mailing list instead, as the change is not GitHub specific.

anakryiko:
please submit the patch upstream, thanks

PR #747, opened 2023-11-10
ci: give /dev/kvm 0666 permissions inside CI runner

eddyz87:
Starting recently libbpf CI runs started failing with the following error:

```
    ##[group]vm_init - Starting virtual machine...
    Starting VM with 4 CPUs...
    INFO: /dev/kvm exists
    KVM acceleration can be used
    Could not access KVM kernel module: Permission denied
    qemu-system-x86_64: failed to initialize KVM: Permission denied
    ##[error]Process completed with exit code 2.
```

E.g. see [here](https://github.com/libbpf/libbpf/actions/runs/6819530119/job/18547589967?pr=746). The error happens because CI user has not enough rights to access `/dev/kvm`. On a regular machine the solution would be to add user to group `kvm`, however that would require a re-login, which is cumbersome to achieve in CI setting. Instead, use a recipe described [here](https://stackoverflow.com/questions/37300811/android-studio-dev-kvm-device-permission-denied/61984745#61984745) to make udev set `0666` access permissions for `/dev/kvm`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,20 @@ jobs:
             runs_on: s390x
             arch: 's390x'
     steps:
+      # Allow CI user to access /dev/kvm (via qemu) w/o group change/relogin
+      # by changing permissions set by udev.
+      - name: Set /dev/kvm permissions
+        shell: bash
+        run: |
+          if [ -e /dev/kvm ]; then
+            echo "/dev/kvm exists, updating permissions"
+            echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' \
+              | sudo tee /etc/udev/rules.d/99-kvm4all.rules > /dev/null
+            sudo udevadm control --reload-rules
+            sudo udevadm trigger --name-match=kvm
+          else
+            echo "/dev/kvm does not exist"
+          fi
       - uses: actions/checkout@v3
         name: Checkout
       - uses: ./.github/actions/setup


eddyz87:
Sigh... CI passed for 4.9.0 but failed for other kernels, which is very strange as ultimately there are no differences in setup process. I tried this fix on a toy repository first ([here](https://github.com/eddyz87/actions-test-repo/actions/runs/6825156296/job/18562438486)) and it works there.

eddyz87:
Basing on comments [here](https://github.com/actions/runner-images/discussions/7191#discussioncomment-5189098), it appears that virtualization support is not guaranteed for standard github runners, some VMs have it, some VMs don't. For guaranteed virtualization support either self-hosted or larget github runner is needed.

anakryiko:
Do you think it's possible to teach https://github.com/libbpf/ci/blob/main/run-qemu/run.sh to detect KVM support properly for our case and not require kvm? It still sucks that we'll have random failures depending on which worker Github actions provides?

I'll land this diff, but it doesn't really solve the problem, unfortunately.

eddyz87:
> Do you think it's possible to teach https://github.com/libbpf/ci/blob/main/run-qemu/run.sh to detect KVM support properly for our case and not require kvm?
Sure, that's good idea. In fact, it already has provision for this:

```
if kvm-ok ; then
  accel=$kvm_accel
else
  accel=$tcg_accel
fi
```

But for some reason this check is not sufficient. I'll take a look.

> It still sucks that we'll have random failures depending on which worker Github actions provides?
Yeap, and we start 4 runners, so probability of getting lucky on all 4 is low.

eddyz87:
> I'll land this diff, but it doesn't really solve the problem, unfortunately.

Let's wait with landing this. The above `kvm-ok` check should have saved us.

anakryiko:
> > I'll land this diff, but it doesn't really solve the problem, unfortunately.
> 
> Let's wait with landing this. The above `kvm-ok` check should have saved us.

ok, I'll wait, thanks for taking a look!

chantra:
I think `kvm-ok` is a mere check that succeed if /dev/kvm exists (regardless of permissions).

It seems GH has been slowly rolling out nested kvm to the GH hosted runners. So I will expect this to be the new normal eventually.

GH large runners already offer nested virtualization and their recommended approach is:

https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/

I think we want to make sure we only apply this to the non-self-hosted runners though, or at least make sure it does not break that use case.

eddyz87:
> I think `kvm-ok` is a mere check that succeed if /dev/kvm exists (regardless of permissions).
> 
> It seems GH has been slowly rolling out nested kvm to the GH hosted runners. So I will expect this to be the new normal eventually.

That might be the case. This would also explain extreme slowness of the libbpf CI, if it was using non-accelerated QEMU.

> GH large runners already offer nested virtualization and their recommended approach is:
> 
> https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/

Yes, that's exactly the trick I use in the MR.
 
> I think we want to make sure we only apply this to the non-self-hosted runners though, or at least make sure it does not break that use case.

I'll update the MR to apply udev conditionally, only if /dev/kvm exists.

eddyz87:
> I think `kvm-ok` is a mere check that succeed if /dev/kvm exists (regardless of permissions).

You are correct, I checked source code from [here](https://launchpad.net/ubuntu/+source/cpu-checker/0.7-1.1). Turns out `kvm-ok` is a small bash script and relevant part looks as follows:

```
...
# Now, check that the device exists
if [ -e /dev/kvm ]; then
    echo "INFO: /dev/kvm exists"
    verdict 0
else
    echo "INFO: /dev/kvm does not exist"
    echo "HINT:   sudo modprobe kvm_$brand"
fi
...
```

Thank you for suggesting this.

eddyz87:
`/dev/kvm` permission changes are now guarded by `if` that checks for `/dev/kvm` existence.
Example of a run on the runner with kvm: [here](https://github.com/eddyz87/libbpf/actions/runs/6830057882/job/18578064849).
Example of a run on the runner without kvm: [here](https://github.com/eddyz87/libbpf/actions/runs/6830057882/job/18578028992).
(Canceled both once qemu started executing).

Still it appears s390x runner does not like udev changes (and it is a self-hosted runner).

anakryiko:
thanks, @eddyz87! Can you please take care of failure on 5.5? Latest on x86-64 is failing to boot in qemu, could it be because of these changes?

eddyz87:
> thanks, @eddyz87! Can you please take care of failure on 5.5? Latest on x86-64 is failing to boot in qemu, could it be because of these changes?

I'll look into it.

Issue #754, opened 2023-11-17
CI: find_vma test fails on LATEST kernel


eddyz87:
Something is wrong with find_vma test on libbpf CI on LATEST kernel and I can't figure out what.
This test has a loop that waits for a few variables to be set by BPF program `SEC("perf_event") int handle_pe(void)`.
However, these variables are never set, log looks as follows:

```
    serial_test_find_vma:PASS:find_vma__open_and_load 0 nsec
    test_find_vma_pe:PASS:attach_perf_event 0 nsec
    test_and_reset_skel:FAIL:found_vm_exec unexpected found_vm_exec: actual 0 != expected 1
    test_and_reset_skel:FAIL:find_addr_ret unexpected find_addr_ret: actual -1 != expected 0
    test_and_reset_skel:FAIL:find_zero_ret unexpected find_zero_ret:
```

Relevant parts of the `prog_tests/find_vma.c`:

```c
static void test_and_reset_skel(struct find_vma *skel, int expected_find_zero_ret, bool need_test)
{
    if (need_test) {
        ASSERT_EQ(skel->bss->found_vm_exec, 1, "found_vm_exec");
        ASSERT_EQ(skel->data->find_addr_ret, 0, "find_addr_ret");
        ASSERT_EQ(skel->data->find_zero_ret, expected_find_zero_ret, "find_zero_ret");
        ASSERT_OK_PTR(strstr(skel->bss->d_iname, "test_progs"), "find_test_progs");
    }
    ...
}
...
static bool find_vma_pe_condition(struct find_vma *skel)
{
    return skel->bss->found_vm_exec == 0 ||
        skel->data->find_addr_ret != 0 ||
        skel->data->find_zero_ret == -1 ||
        strcmp(skel->bss->d_iname, "test_progs") != 0;
}
...
static void test_find_vma_pe(struct find_vma *skel)
{
    const int one_bn = 1000000000;
    ...
    pfd = open_pe();
    ...
    for (i = 0; i < one_bn && find_vma_pe_condition(skel); ++i)
        ++j;

    test_and_reset_skel(skel, -EBUSY /* in nmi, irq_work is busy */, i == one_bn);
                                                                  // ^^^
                                                                  // this is true on CI
cleanup:
    bpf_link__destroy(link);
    close(pfd);
}
```

[...]

PR #755, opened 2023-11-21
ci: disable selftests failing on LATEST and 5.5 kernels

eddyz87:
This test started to fail on LATEST kernel, e.g. see [0]. There are no apparent issues with it on main kernel CI. Link [1] has some details. Overall, it looks like perf event expected by sub-test test_find_vma_pe() of
prog_tests/find_vma.c is never delivered.

[0] https://github.com/libbpf/libbpf/actions/runs/6904382140/job/18784917164
[1] https://github.com/libbpf/libbpf/issues/754

### DIFF
--- a/ci/vmtest/configs/ALLOWLIST-5.5.0
+++ b/ci/vmtest/configs/ALLOWLIST-5.5.0
@@ -33,7 +33,6 @@ rdonly_maps
 section_names
 signal_pending
 sockmap_ktls
-sockopt
 spinlock
 stacktrace_map
 stacktrace_map_raw_tp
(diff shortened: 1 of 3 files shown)


anakryiko:
let's also ban bpf_cookie/perf_event on LATEST and sockopt on 5.5? The first seems to now detecting hardware event support, which actually doesn't work. The latter might be some new subtests that were added recently and expect new kernel features or something?

eddyz87:
Ok, will aggregate all denylist changes as a single MR.

eddyz87:
Split in two commits, added details in each commit message.

anakryiko:
Thanks a lot, landing.

anakryiko:
actually, it's marked as draft, can you please mark it as ready for review?

PR #758, opened 2023-12-05
ci: custom patch to patch out BPF_F_TEST_REG_INVARIANTS flag

anakryiko:
Without needing to modify tons of BPF selftests file, make sure we don't pass BPF_F_TEST_REG_INVARIANTS to kernel, to make BPF selftests work on 4.9 and 5.5 kernels.

Also undo previously denylisted tests, assumed to be due to perf events.

Thank you for considering a contribution!

Please note that the `libbpf` authoritative source code is developed as part of bpf-next Linux source tree under tools/lib/bpf subdirectory and is periodically synced to Github. As such, all the libbpf changes should be sent to BPF mailing list, please don't open PRs here unless you are changing Github-specific parts of libbpf (e.g., Github-specific Makefile).


eddyz87 (approved):

eddyz87:
All makes sense.
Looks like two my older commits are still necessary, as test disabled in these commits don't use `bpf_prog_test_load` or `test_loader.c`:
a6b990991c4c ("ci: disable sockopt selftest for 5.5 kernel")
4161e1f41d9a ("ci: disable a number of selftest causing CI for LATEST kernel")

PR #761, opened 2023-12-13
ci: move /dev/kvm permissions setup from to actions/vmtest.yml

eddyz87:
The vmtest action is used from by workflows: test, pahole, ondemand. At the same time vmtest action requires valid access rights to /dev/kvm and is the only action that uses it.
This commit moves /dev/kvm permissions setup from test workflow to vmtest action, in order to make sure that setup logic is shared by all workflows that run vmtest.
Should fix CI failures like [1].

[1] https://github.com/libbpf/libbpf/actions/runs/7104762048/job/19340484589

### DIFF
--- a/.github/actions/vmtest/action.yml
+++ b/.github/actions/vmtest/action.yml
@@ -16,6 +16,22 @@ inputs:
 runs:
   using: "composite"
   steps:
+    # Allow CI user to access /dev/kvm (via qemu) w/o group change/relogin
+    # by changing permissions set by udev.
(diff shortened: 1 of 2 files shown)


eddyz87:
I tried this change with pahole workflow executed on `pull_request`, it finished ok (after re-run, `#164/2   perf_branches/perf_branches_no_hw:FAIL` is a flip-flop test). Here is a link to the run:

https://github.com/libbpf/libbpf/actions/runs/7199960281

anakryiko:
great, thank you!

PR #768, opened 2024-01-26
ci: seems like LLVM 19 is now the latest one

anakryiko:
Adapt version checks to take LLVM 19 into account.


anakryiko:
Not broken, not going to fix it.

PR #800, opened 2024-04-25
sync: Sync .mailmap entries

qmonnet:
The kernel repository has a .mailmap file to remap author names and email addresses to their desired format in Git logs (for details, see [gitmailmap documentation][0]). Alas, this is only visible for author information when looking at the logs locally, as GitHub [does not support mailmaps][1] at the moment.

This commit adds a .mailmap file for libbpf, automatically generated from the kernel's version. The script to generate the .mailmap is added, too: it works by grepping email addresses from authors in the repository, and collecting all lines ending with this address in the kernel's .mailmap - in other words, all lines where this address is used as a pattern for a remapping.

To keep the .mailmap up-to-date, add a call to the script to sync-kernel.sh.

[0]: https://git-scm.com/docs/gitmailmap
[1]: https://github.com/orgs/community/discussions/22518

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,16 @@
+Alexei Starovoitov <ast@kernel.org> <alexei.starovoitov@gmail.com>
+Antoine Tenart <atenart@kernel.org> <antoine.tenart@bootlin.com>
+Björn Töpel <bjorn@kernel.org> <bjorn.topel@intel.com>
+Changbin Du <changbin.du@intel.com> <changbin.du@gmail.com>
+Colin Ian King <colin.i.king@gmail.com> <colin.king@canonical.com>
+Dan Carpenter <error27@gmail.com> <dan.carpenter@oracle.com>
+Geliang Tang <geliang@kernel.org> <geliang.tang@suse.com>
+Herbert Xu <herbert@gondor.apana.org.au>
+Jakub Kicinski <kuba@kernel.org> <jakub.kicinski@netronome.com>
+Leo Yan <leo.yan@linux.dev> <leo.yan@linaro.org>
+Mark Starovoytov <mstarovo@pm.me> <mstarovoitov@marvell.com>
+Maxim Mikityanskiy <maxtram95@gmail.com> <maximmi@mellanox.com>
+Maxim Mikityanskiy <maxtram95@gmail.com> <maximmi@nvidia.com>
+Quentin Monnet <qmo@kernel.org> <quentin@isovalent.com>
+Quentin Monnet <qmo@kernel.org> <quentin.monnet@netronome.com>
+Vadim Fedorenko <vadim.fedorenko@linux.dev> <vfedorenko@novek.ru>
--- a/scripts/mailmap-update.sh
+++ b/scripts/mailmap-update.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -eu
+
+usage () {
+    echo "USAGE: ./mailmap-update.sh <libbpf-repo> <linux-repo>"
+    exit 1
+}
+
+LIBBPF_REPO="${1-""}"
+LINUX_REPO="${2-""}"
+
+if [ -z "${LIBBPF_REPO}" ] || [ -z "${LINUX_REPO}" ]; then
+    echo "Error: libbpf or linux repos are not specified"
+    usage
+fi
+
+LIBBPF_MAILMAP="${LIBBPF_REPO}/.mailmap"
+LINUX_MAILMAP="${LINUX_REPO}/.mailmap"
+
+tmpfile="$(mktemp)"
+cleanup() {
+    rm -f "${tmpfile}"
+}
+trap cleanup EXIT
+
+grep_lines() {
+    local pattern="$1"
+    local file="$2"
+    grep "${pattern}" "${file}" || true
+}
+
+while read -r email; do
+    grep_lines "${email}$" "${LINUX_MAILMAP}" >> "${tmpfile}"
+done < <(git log --format='<%ae>' | sort -u)
+
+sort -u "${tmpfile}" > "${LIBBPF_MAILMAP}"
--- a/scripts/sync-kernel.sh
+++ b/scripts/sync-kernel.sh
@@ -352,4 +352,10 @@ else
 	esac
 fi
 
+echo "Regenerating .mailmap..."
+cd_to "${LINUX_REPO}"
+git checkout "${TIP_SYM_REF}"
+cd_to "${LIBBPF_REPO}"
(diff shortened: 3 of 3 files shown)


.mailmap
```diff
@@ -0,0 +1,16 @@
...
+Dan Carpenter <error27@gmail.com> <dan.carpenter@oracle.com>
+Geliang Tang <geliang@kernel.org> <geliang.tang@suse.com>
+Herbert Xu <herbert@gondor.apana.org.au>
+Jakub Kicinski <kuba@kernel.org> <jakub.kicinski@netronome.com>
+Leo Yan <leo.yan@linux.dev> <leo.yan@linaro.org>
+Mark Starovoytov <mstarovo@pm.me> <mstarovoitov@marvell.com>
+Maxim Mikityanskiy <maxtram95@gmail.com> <maximmi@mellanox.com>
+Maxim Mikityanskiy <maxtram95@gmail.com> <maximmi@nvidia.com>
+Quentin Monnet <qmo@kernel.org> <quentin@isovalent.com>
+Quentin Monnet <qmo@kernel.org> <quentin.monnet@netronome.com>
```
anakryiko, line 15: can entries like this be combined in mailmap?

scripts/mailmap-update.sh
```diff
@@ -0,0 +1,37 @@
...
+
+grep_lines() {
+    local pattern="$1"
+    local file="$2"
+    grep "${pattern}" "${file}" || true
+}
+
+while read -r email; do
+    grep_lines "${email}$" "${LINUX_MAILMAP}" >> "${tmpfile}"
+done < <(git log --format='<%ae>' | sort -u)
```
anakryiko, line 35: is this going to do a full history log? that's quite slow. Maybe let's log only newly synced commits, and then merge new authors and those that are in .mailmap? That would be much faster, right?

anakryiko (changes requested):

.mailmap
qmonnet: I believe they can't :confused:

scripts/mailmap-update.sh
qmonnet: We can, although the operation doesn't seem to take that long on my laptop:

```console
$ time (git log --format='<%ae>' | sort -u > /dev/null)
( git log --format='<%ae>' | sort -u > /dev/null; )  0.05s user 0.01s system 118% cpu 0.047 total
```

I can look into it in the coming days.

scripts/mailmap-update.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -eu
+
+usage () {
+    echo "USAGE: ./update-mailmap.sh <bpftool-repo> <linux-repo>"
```
anakryiko, line 6: mailmap-update.sh ?

scripts/mailmap-update.sh
anakryiko: ah, this is supposed to be run from libbpf repo, not linux repo, right? I see, ok, let's keep it simple for now.

scripts/sync-kernel.sh
```diff
@@ -352,4 +352,10 @@ else
 	esac
 fi
 
+echo "Regenerating .mailmap..."
+cd_to "${LINUX_REPO}"
+git checkout "${TIP_SYM_REF}"
+cd_to .
```
anakryiko, line 358: let's do `cd_to "${LIBBPF_REPO}"`, `.` is workdir which can be any dir

anakryiko (changes requested):

scripts/mailmap-update.sh
qmonnet: Indeed!

qmonnet: It even says "bpftool" instead of "libbpf" :facepalm:

qmonnet:
Thanks for the review, your comments should be addressed.

<details><summary>Incremental diff</summary>

```diff
diff --git a/scripts/mailmap-update.sh b/scripts/mailmap-update.sh
index a6d7c1d0722f..13a3dcf3f65e 100755
--- a/scripts/mailmap-update.sh
+++ b/scripts/mailmap-update.sh
@@ -3,7 +3,7 @@
 set -eu
 
 usage () {
-    echo "USAGE: ./update-mailmap.sh <bpftool-repo> <linux-repo>"
+    echo "USAGE: ./mailmap-update.sh <libbpf-repo> <linux-repo>"
     exit 1
 }
 
diff --git a/scripts/sync-kernel.sh b/scripts/sync-kernel.sh
index 74783571425d..36bb96f24c41 100755
--- a/scripts/sync-kernel.sh
+++ b/scripts/sync-kernel.sh
@@ -355,7 +355,7 @@ fi
 echo "Regenerating .mailmap..."
 cd_to "${LINUX_REPO}"
 git checkout "${TIP_SYM_REF}"
-cd_to .
+cd_to "${LIBBPF_REPO}"
 "${LIBBPF_REPO}"/scripts/mailmap-update.sh "${LIBBPF_REPO}" "${LINUX_REPO}"
 
 cleanup
```
</details>

anakryiko (approved):
Looks great, thanks a lot!

anakryiko:
@qmonnet one thing that became obvious doing a sync with new .mailmap logic is that sync script doesn't add a new commit if there are .mailmap changes. Can you please help adding this check/step to the script, so that no one has to do this manually? Thanks!

qmonnet:
Ah right, forgot the commit part :facepalm:. Sure, I'll add it

scripts/mailmap-update.sh
qmonnet: Thinking again about this: If we were to do incremental updates, we could miss new entries in Linux's mailmap.

For example:

- Someone, contributor C, contributed their changes in libbpf v1.4
- Contributor C does not send new contributions to libbpf after v1.4 has been tagged
- We get more contributions, sync the repo...
- Meanwhile, contributor C adds an entry for their email in Linux's .mailmap
- We're updating libbpf's .mailmap, considering only the contributions since the latest sync, say since v1.4: There's no contributions from contributor C on that period, so we're not looking for their email in Linux's .mailmap. We're not adding any entry (although one exists on Linux's side) and we don't reflect the email or name change in `git log`.

anakryiko: Yes, but libbpf log keeps growing over time and if it starts taking more time just to list all commits we'd need to think about this problem some more. For now it's not a big deal, so let's not fix what's not broken.

qmonnet: Agreed. We should be good for some time: the command takes me 0.3s in Cilium's repo with its 28k+ commits.

PR #805, opened 2024-05-01
sync: Commit .mailmap changes from script when sync-ing repo

qmonnet:
In commit 4794f18bf468 ("sync: Sync .mailmap entries"), we updated the sync-up script to automatically update libbpf's .mailmap; however, the script would not take care of committing the changes. Let's address this.

The code is copied and adapted from the part where we commit changes to src/bpf_helper_defs.h.

Follow-up to https://github.com/libbpf/libbpf/pull/800#issuecomment-2085813085.

### DIFF
--- a/scripts/sync-kernel.sh
+++ b/scripts/sync-kernel.sh
@@ -295,6 +295,22 @@ Latest changes to BPF helper definitions.
 " -- src/bpf_helper_defs.h
 fi
 
+echo "Regenerating .mailmap..."
+cd_to "${LINUX_REPO}"
+git checkout "${TIP_SYM_REF}"
+cd_to "${LIBBPF_REPO}"
+"${LIBBPF_REPO}"/scripts/mailmap-update.sh "${LIBBPF_REPO}" "${LINUX_REPO}"
+# if anything changed, commit it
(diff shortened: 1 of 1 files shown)


scripts/sync-kernel.sh
```diff
@@ -357,5 +357,15 @@ cd_to "${LINUX_REPO}"
 git checkout "${TIP_SYM_REF}"
 cd_to "${LIBBPF_REPO}"
 "${LIBBPF_REPO}"/scripts/mailmap-update.sh "${LIBBPF_REPO}" "${LINUX_REPO}"
+# if anything changed, commit it
```
anakryiko, line 360: looks good, but let's move entire .mailmap step to right after bpf_helpers_defs.h, so that "cover letter" commit includes it as well in the summary?

qmonnet: Sounds good, I moved it. The new commit will be “before” the cover letter, but will not be listed in the summary, which is generated before that.

anakryiko: Ah, ok, seems like we don't put bpf_helper_defs.h into cover letter as well, that's fine. But I think this order of commits is more logical, thanks!

PR #811, opened 2024-05-29
ci: switch to ubuntu-latest where possible

anakryiko:


chantra:
Just to know, what is the reasoning to move to latest?
FYI, I plan to update kernel/parches/vmtest to 22.04 in the coming months more for a somewhat up-to-date base and newer gcc.
As much as getting libbpf fully on "latest" we could either provide a new rootfs which uses the same Ubuntu version, or move to vmtest (which may be relatively easy).

chantra:
I am assuming it is toolchain related, but just want to make sure.

anakryiko:
> I am assuming it is toolchain related, but just want to make sure.

s390x build test started failing claiming libelf can't be found (and thus `#include <libelf.h>` couldn't be satisfied as well). Unfortunately I still can't move to ubuntu-latest for BPF selftests tests due to GLIBC version restrictions, so updating QEMU image would definitely help.

PR #844, opened 2024-08-22
ci: run test on Ubuntu 24.04

chantra:
Migrate builds to use Ubuntu 24.04 runners.
vmtest will need to run on Ubuntu 20.04 rootfs too (see https://github.com/libbpf/ci/pull/136)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,16 +19,16 @@ jobs:
       matrix:
         include:
           - kernel: 'LATEST'
-            runs_on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


chantra:
This would need to land after https://github.com/libbpf/ci/pull/136 so the new Ubuntu 24.04 rootfs can be used.

chantra:
newly forced pushed commit will break until https://github.com/libbpf/ci/pull/136 is landed.

chantra:
landed now that libbpf/ci#136 is also landed.

PR #853, opened 2024-10-07
ci: get rid of s390x kernel tests

anakryiko:
Kernel/libbpf code is very well tested on s390x in BPF CI, so get rid of it here as it often is a source of trouble and noise, without really benefiting us much.


theihor:
Closing in favor of #854

PR #862, opened 2024-11-01
[ci] use libbpf/ci/run-vmtest action to run selftests

theihor:
This is a "stage 1" of reconciling libbpf CI workflows with BPF CI.

Depends on: https://github.com/libbpf/ci/pull/150

#### Summary

Local vmtest action is refactored into a local reusable workflow.

Local build_selftests action is replaced with libbpf/ci/build-selftestes@v2
I left hard-committed `vmlinux.h` at `.github/actions/build_selftests/vmlinux.h`, because I was concerned it may break the sync routines; let me know if it's safe to move.

prepare-rootfs + run-qemu steps are replaced with libbpf/ci/run-vmtest@v2

pahole_staging workflow is removed, it is now covered by setting another possible value ("tmp.master") for `pahole` input to reusable workflow

Sometimes a little of bash is necessary to glue things together. In such cases, I think we should prefer local scripts to inline, as it is easier to debug locally. An example of a script like that is `ci/vmtest/prepare-selftests-run.sh`.

Old temp diffs are removed.


theihor:
@anakryiko @chantra this one is ready for review, but it shouldn't be merged before https://github.com/libbpf/ci/pull/150

Next step is to improve the build routines (libbpf/ci/build-linux will probably need changes), so that the kbuild result is cached and other linux versions (4.19, 5.10 etc.) are relatively easy to add.

.github/workflows/test.yml
```diff
@@ -1,33 +1,41 @@
...
-        with:
-          kernel: ${{ matrix.kernel }}
-          arch: ${{ matrix.arch }}
+            llvm-version: '18'
+            pahole: 'master'
+          - kernel: 'LATEST'
+            runs_on: 'ubuntu-24.04'
+            arch: 'x86_64'
+            llvm-version: '17'
+            pahole: 'master'
```
anakryiko, line 28: do we need two LATEST with clang 17 and clang 18?

.github/workflows/vmtest.yml
```diff
@@ -0,0 +1,136 @@
...
+          patches-root: '${{ github.workspace }}/ci/diffs'
+          repo-root: '.kernel'
+
+      - name: Configure kernel build
+        shell: bash
+        run: |
+          cd .kernel
+          cat tools/testing/selftests/bpf/config \
+              tools/testing/selftests/bpf/config.${{ inputs.arch }} > .config
+          # this file might or mihgt not exist depending on kernel version
```
anakryiko, line 70: typo: might

anakryiko (approved):
Very nice, thank you!

Please let me know when we can actually land this.

.github/workflows/test.yml
theihor: That's a question to you. On BPF CI we test both 17 and 18. Before this PR libbpf CI tested only version 17.

Let me know what do we want tested.

theihor:
> Please let me know when we can actually land this.

We can now land this.

PR #880, opened 2025-01-18
libbpf.c: bpf_program__attach_uprobe_opts fix possible memory leaks.

devnexen:
bpf_program__attach_perf_event_opts() might be not enough to close the file descriptor, bpt_link__destroy() does a more thorough clean up including its inner file descriptor.


anakryiko:
Libbpf's authoritative source code is part of Linux kernel, so this has to be submitted through bpf@vger.kernel.org mailing list, we can't accept it through Github, unfortunately.

devnexen:
gotcha I ll do it then.

PR #881, opened 2025-01-23
ci: fix build failures on s390x and ppc64le

theihor:


theihor:
Closing in favor of #882

PR #882, opened 2025-01-31
ci: remove dependency on run-on-arch-action

theihor:
run-on-arch-action is simply a wrapper around docker. There is no value in using it in libbpf, as it is not complicated to run non-native arch docker images directly on github-hosted runners.

Docker relies on qemu-user-static installed on the system to emulate different architectures.

Recently there were various reports about multi-arch docker builds failing with seemingly random issues, and it appears to boil down to qemu [1]. I stumbled on this problem while updating s390x runners [2] for BPF CI, and setting up more recent version of qemu helped.

This change addresses recent build failures on s390x and ppc64le.

[1] https://github.com/docker/setup-qemu-action/issues/188
[2] https://github.com/kernel-patches/runner/pull/69
[3] https://docs.docker.com/build/buildkit/#getting-started

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,31 +61,32 @@ jobs:
           - arch: aarch64
           - arch: ppc64le
           - arch: s390x
-          - arch: x86
+          - arch: amd64
     steps:
       - uses: actions/checkout@v4
         name: Checkout
+
+      - name: Setup QEMU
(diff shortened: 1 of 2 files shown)


theihor:
More details here: https://github.com/kernel-patches/runner/pull/67#issuecomment-2628388670

PR #892, opened 2025-03-31
libbpf: add proto_func param name generation on sanitizing it to enum…

T3RR7:
Recently I've encountered some issues with loading on debian 10, kernel 4.19.0-[17-22].

This PR is a fix for one of those. When the kernel loads BTF with specified min-CORE and libbpf does some sanitizing on those it "translates" `func_proto` to `enum`, but if `func_proto` has arguments for it's parameters then kernel verifier fails with "Invalid name" error. This error caused because `enum` members [*must* has a valid C identifier name](https://github.com/torvalds/linux/blob/609706855d90bcab6080ba2cd030b9af322a1f0c/kernel/bpf/btf.c#L4350).

This PR proposes to generate this name during sanitation of `func_proto`


anakryiko:
https://github.com/libbpf/libbpf?tab=readme-ov-file#libbpf-and-general-bpf-usage-questions

PR #893, opened 2025-03-31
libbpf: add check if kernel supports kind flag and fix the bitfield m…

T3RR7:
The second issue in addition to https://github.com/libbpf/libbpf/pull/892 is a kind flag.

Let's say we're compiling the BPF object on new kernel with new compiler. It's with 99% percent chance that compiler will use this `kind flag` ([docs](https://github.com/torvalds/linux/blob/v4.19/kernel/bpf/btf.c#L166)) which [changes the behavior of bitfield members relocation](https://docs.kernel.org/bpf/btf.html#btf-kind-struct).

But of we're loading on old kernels that have no support of this it will fail in multiple places
- The type metadata check ([BTF_INFO_MASK prohibits 31d bit](https://github.com/torvalds/linux/blob/v4.19/kernel/bpf/btf.c#L166))
- [Integral bitfield verification](https://github.com/torvalds/linux/blob/609706855d90bcab6080ba2cd030b9af322a1f0c/kernel/bpf/btf.c#L2133). Because kernel waits for a offset-only encoding.

What this PR proposes:
- The feature check whether kernel supports the kind flag or not
- Kind flag sanitizing 
- Struct bitfield members sanitizing by generation a proper replacement the type of bitfield with corresponding integer type  

I assume that there is a more elegant way to do it. Maybe it could be better to write an API like `btf_int` like in kernel to avoid code like [this](https://github.com/libbpf/libbpf/compare/master...T3RR7:libbpf:kind-flag-support-check-and-fix?expand=1#diff-ac82747bd450e8e5e0805a99d6c24afa9fcbfcddffbce5aa3288b2edabe17419R3275). Anyway, it works.

### DIFF
--- a/src/features.c
+++ b/src/features.c
@@ -507,6 +507,33 @@ static int probe_kern_arg_ctx_tag(int token_fd)
 	return probe_fd(prog_fd);
 }
 
+static int probe_kern_btf_type_kind_flag(int token_fd)
+{
+    const char strs[] = "\0bpf_spin_lock\0val\0cnt\0l";
+    /* struct bpf_spin_lock {
+     *   int val;
+     * };
+     * struct val {
+     *   int cnt;
+     *   struct bpf_spin_lock l;
+     * };
+     */
+    __u32 types[] = {
+        /* int */
+        BTF_TYPE_INT_ENC(0, BTF_INT_SIGNED, 0, 32, 4),  /* [1] */
+        /* struct bpf_spin_lock */                      /* [2] */
+        BTF_TYPE_ENC(1, BTF_INFO_ENC(BTF_KIND_STRUCT, 1 /* kind bit */, 1), 4),
+        BTF_MEMBER_ENC(15, 1, 0), /* int val; */
+        /* struct val */                                /* [3] */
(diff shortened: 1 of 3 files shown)


anakryiko:
You've already sent patches upstream (as per https://github.com/libbpf/libbpf?tab=readme-ov-file#libbpf-and-general-bpf-usage-questions), will close these PRs, this repo is a mirror, we can't land libbpf code through here.

T3RR7:
> You've already sent patches upstream (as per https://github.com/libbpf/libbpf?tab=readme-ov-file#libbpf-and-general-bpf-usage-questions), will close these PRs, this repo is a mirror, we can't land libbpf code through here.

Yep, but here is a more convenient place to discuss proposed changes, what do you think?

If you think no then I'll wait for your reply at the mailing list :)

anakryiko:
> Yep, but here is a more convenient place to discuss proposed changes, what do you think?
> 

Convenient? Yes, probably. But no one from upstream is looking at these issues besides me. So no, please follow the readme and keep conversation on the mailing list, please.

PR #919, opened 2025-08-23
sync: uapi: move constants from <linux/kernel.h> to <linux/const.h>

yuyuyureka:
This change was applied to the kernel tree in 2020 afaict, to fix redefinition issues with musl-libc.

I ran into this issue while building with the headers included with libbpf, since it seems they were not synced since then.

Signed-off-by: Yureka Lilian <yuka@yuka.dev>

### DIFF
--- a/include/uapi/linux/netlink.h
+++ b/include/uapi/linux/netlink.h
@@ -2,7 +2,7 @@
 #ifndef __LINUX_NETLINK_H
 #define __LINUX_NETLINK_H
 
-#include <linux/kernel.h>
+#include <linux/const.h>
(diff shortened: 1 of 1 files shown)


anakryiko:
@yuyuyureka thanks for bringing this discrepancy to my attention! It turned out this is not the only UAPI header difference there is, so I went ahead and fixed up sync scripts and synced a bunch of UAPI headers, including this one (but it adds more stuff beyond just a single line change you did).

I hope you don't mind that https://github.com/libbpf/libbpf/pull/920 will supersede your change. I'm going ahead and close this one for now.

PR #975, opened 2026-06-05
ci: denylist flaky 'file_reader' selftest

theihor:
file_reader/on_open_expect_fault expects MADV_PAGEOUT to evict the file page so a non-sleepable bpf_dynptr_read() returns -EFAULT. In the CI VM the page is still resident at read time, the read succeeds, run_success stays 0 and the subtest fails on every nightly run (all three flavors). Already denied in the kernel-patches CI; mirror it in the bundled denylist so libbpf-ci goes green again.


eddyz87:
Should we ping Mykyta to fix the test?

theihor:
Oh, actually there is a patch: https://lore.kernel.org/all/20260603-file_reader_flake-v1-1-7f3f52d1e388@meta.com/

Let me try that.

